PR #4, opened 2020-07-28
Witnesses for certs and withdrawals

PeterBenc:
Return witnesses also for certificates and withdrawals, with unique paths.

### DIFF
--- a/src/Ada.js
+++ b/src/Ada.js
@@ -658,6 +658,17 @@ export default class Ada {
 
     // init
     //console.log("init");
+    
+    // for unique witness paths
+    const witnessPathsSet = new Set();
+    const witnessPaths = [];
+    for (const {path} of [...inputs, ...certificates, ...withdrawals]) {
+      const pathKey = JSON.stringify(path);
+      if (!witnessPathsSet.has(pathKey)) {
+        witnessPathsSet.add(pathKey);
+        witnessPaths.push(path);
+      }
+    }
     await signTx_init(
       networkId,
       protocolMagic,
@@ -665,7 +676,7 @@ export default class Ada {
       outputs.length,
       certificates.length,
       withdrawals.length,
-      inputs.length + certificates.length + withdrawals.length,
+      witnessPaths.length,
       metadataHashHex != null
     )
     // inputs
@@ -726,7 +737,7 @@ export default class Ada {
 
     //console.log("witnesses");
     const witnesses = [];
-    for (const {path} of [...inputs, ...certificates, ...withdrawals]) {
+    for (const path of witnessPaths) {
       const witness = await signTx_getWitness(path);
       witnesses.push(witness);
     }
--- a/src/Ada.js
+++ b/src/Ada.js
@@ -665,7 +665,7 @@ export default class Ada {
       outputs.length,
       certificates.length,
       withdrawals.length,
-      inputs.length,
+      inputs.length + certificates.length + withdrawals.length,
       metadataHashHex != null
     )
     // inputs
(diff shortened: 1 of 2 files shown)


refi93:
did you run the integration tests and do they pass? I would add a test to verify that we are supplying also certificate/withrawals - should be quick

src/Ada.js
```diff
@@ -725,11 +725,16 @@ export default class Ada {
     const { txHashHex } = await signTx_awaitConfirm();
 
     //console.log("witnesses");
+    const uniquePaths = new Set()
```
refi93, line 728: add semicolon for consistency with the rest of code

src/Ada.js
```diff
@@ -725,11 +725,16 @@ export default class Ada {
...
-    for (const input of inputs) {
-      const witness = await signTx_getWitness(input.path);
+    for (const {path} of [...inputs, ...certificates, ...withdrawals]) {
+      const pathKey = JSON.stringify(path)
+      if (uniquePaths.has(pathKey)) continue;
+      uniquePaths.add(pathKey)
+      const witness = await signTx_getWitness(path);
       witnesses.push(witness);
     }
+
```
janmazak, line 737: The number of witnesses has to be announced to Ledger in advance (in the initial APDU). If the number is only determined here, Ledger will not accept these (superfluous) witness requests.

In particular, in line 668, an incorrect argument is passed for numWitnesses.

janmazak: So my suggestion would be to first construct the list of required witness paths, then pass its size to `signTx_init`, and then process the list here.

PeterBenc:
Fixed failing tests and added a new test for de-duplicating witnesses.

src/Ada.js
```diff
@@ -658,14 +658,18 @@ export default class Ada {
 
     // init
     //console.log("init");
+    const uniquePaths = new Set();
```
janmazak, line 661: Since the scope of this variable is rather large, I would use a name that clarifies that this in only used in dealing with witnesses, e.g. `uniqueWitnessPaths`. 

Even better would be to construct the list of unique paths here and refer to it everywhere we need it. Then we can drop `unique` from the name, and have a `witnessPaths` list. A one-line explanatory comment about uniqueness here would be then sufficient to deal with it once and for all --- the uniqueness does not affect usages of this variable.

If we compute the unique paths in two places, it is somewhat fragile: in the future, it is easy to change it in one place and forget to change the other.

src/Ada.js
```diff
@@ -658,14 +658,18 @@ export default class Ada {
...
+    for (const {path} of [...inputs, ...certificates, ...withdrawals]) {
+      uniquePaths.add(JSON.stringify(path))
```
janmazak, line 663: ;

src/Ada.js
PeterBenc: Yes, but what I need in the first case of computing`witnessPaths`, is their length, in the second case I don't need them at all. All I do is use Set properties as a way to filter the unique paths. Can't really see a way to merge this functionality into something re-usable. Open to suggestions.

src/Ada.js
```diff
@@ -658,14 +658,20 @@ export default class Ada {
 
     // init
     //console.log("init");
+    
+    // for unique witness paths
+    const witnessPaths = new Set();
```
refi93, line 663: I would name this uniqueWitnessPaths and save a comment

src/Ada.js
```diff
@@ -725,11 +731,16 @@ export default class Ada {
     const { txHashHex } = await signTx_awaitConfirm();
 
     //console.log("witnesses");
+    witnessPaths.clear()
```
refi93, line 734: this seems ungly - I would rather build the list of unique paths above together with the set and use that list here

PeterBenc: just pushed this.

PR #5, opened 2020-08-20
fix test for tx with a withdrawal

PeterBenc:
During testing, I discovered that ledger builds txs with a withdrawal using staking key hash (28 bytes) instead of staking address hash (29 bytes). This PR fixes the test for txs with a withdrawal.


janmazak (approved):

test/src/integration/signTx.js
```diff
@@ -207,17 +207,17 @@ const results = {
     * e1a5d89d92f45fa0001a0d0c25611a002dd2e802182a030a05a1581c1d227aefa4b7731491708
     * 85aadba30aab3127cc611ddbc4999def61c186f
```
refi93, line 208: tx body in the comment is outdated, please update

PeterBenc: Sorry for the delay, done...

refi93 (approved):

PR #12, opened 2020-10-11
Pool registration certificates

janmazak:
Does not fully work yet.


src/Ada.js
```diff
@@ -656,19 +815,48 @@ export default class Ada {
...
-    //console.log("init");
-    
-    // for unique witness paths
-    const witnessPathsSet = new Set();
+    // we need exactly one owner given by path to have a witness
+    let witnessOwner = null;
+    if (isSigningPoolRegistrationAsOwner) {
+      if (certificates.length > 1)
+        throw new Error("A pool registration certificate must be standalone");
+      if (withdrawals.length > 1)
```
xdzurman, line 823: should be either withdrawals.length > 0 or withdrawals.length >= 1 or just withdrawals.length

xdzurman (changes requested):

janmazak:
Should fully work now, apart from missing tests.

What about the too long address? I have replaced it in `poolCertificate.js`, but not in `integration/signTx.js`.

src/Ada.js
```diff
@@ -65,10 +65,58 @@ export type StakingBlockchainPointer = {|
...
+|}
+
+export type Margin = {|
+  numeratorStr: string,
+  denominatorStr: string,
+|}
+
+export type PoolParams = {|
+  poolKeyHashHex: string,
+  vrfKeyHashHex: string,
```
PeterBenc, line 105: Is this really a hash? From what I can see, this is 32 bytes long in parsed tx. Thus it should be something like `vrfPubKeyHex`...Right? Or is ledger expecting me to hash this?

src/Ada.js
```diff
@@ -65,10 +65,58 @@ export type StakingBlockchainPointer = {|
...
+  pledgeStr: string,
+  costStr: string,
+  margin: Margin,
+  rewardAccountHashHex: string,
```
PeterBenc, line 109: From what I can see, this has 29 bytes in parsed tx. Isnt this supposed to be called something like `rewardAccountAddressHex` ? Or `rewardAddressHex`? Or am I supposed to drop the first byte of the address?

src/Ada.js
refi93: yes, the serialized transaction contains the blake2b-256 hash of the vrf key, even if the vrf key itself is 32 bytes as well. I just checked it and `echo -n aa0888b9da466f04cba88ee70cfc3dfc3df6ef53dea57a94f6762a253b3c21ab | xxd -r -p | b2sum -l 256 -b | awk '{print $1}'` yields the hash included in the sample tx in the hw wallet owner support design specs (https://docs.google.com/document/d/1H4MHj8AysiRg75bVx7xd3PouQWdGfgEDLUqQAn1TJy0/edit#heading=h.60fl10mfqiy9)

src/Ada.js
janmazak: No, it is the reward address. I've renamed the field.

src/Ada.js
PeterBenc: I see, I was automatically assuming its pubkey since its 32 bytes long. Thanks :+1:

src/Ada.js
PeterBenc: Sorry to bother you with this. You renamed it to `rewardAccountHex`. Didn't you mean `rewardAddressHex`?

janmazak: The stupid inconsistent terminology (by IOHK/Cardano) is behind this. Internally, we use reward address for ledger (as it is used in some of their specs). But in most discussions they favor "reward account", in some of the docs, too, so I decided that it would be better to call it account (hopefully avoiding confusion of Daedalus and other wallets).

package.json
```diff
@@ -55,11 +55,13 @@
     "build-example": "bash ./script/build_example.sh",
     "clean-test": "rm -rf test/lib",
     "build-test": "yarn run clean && yarn run build && yarn run clean-test && babel --source-maps -d test/lib test/src && flow-copy-source -v test/src test/lib",
-    "test": "yarn build-test && yarn test-device-run && yarn test-integration-run && yarn test-direct-run",
+    "test": "yarn build-test && yarn test-device-run && yarn test-integration-run && yarn test-direct-run && yarn test-pool-run",
     "test-device": "yarn run build-test && mocha --timeout 3600000 test/lib/device",
     "test-device-run": "mocha --timeout 3600000 test/lib/device",
     "test-integration": "yarn run build-test && yarn run test-integration-run",
     "test-integration-run": "mocha --timeout 600000 test/lib/integration",
+    "test-pool": "yarn run build-test && yarn run test-pool-run",
```
refi93, line 63: pools should not have its own target in the final version and even now you can quite conveniently grep out the stakepool registration related tests by name, see comment below

```
"//": "run single test by specifying --grep <name> parameter in test-integration"
```

src/Ada.js
```diff
@@ -522,6 +593,9 @@ export default class Ada {
     const signTx_addInput = async (
       input: InputTypeUTxO
     ): Promise<void> => {
+      Precondition.checkIsHexString(input.txHashHex, "invalid tx hash");
+      Precondition.check(input.txHashHex.length == 32 * 2, "invalid tx hash");
```
refi93, line 597: this should be `===`, otherwise it's unnecessary susceptible for edge cases like '64' == 32*2 being true and similar as type conversion is happening in the background

refi93: I see `==` being used in several other places not necessarily touched by this PR, so I would recommend refactoring those to `===` as well

src/cardano.js
```diff
@@ -1,49 +1,198 @@
...
+    Precondition.check(input.txHashHex.length === 32 * 2, "invalid tx hash length");
+
+    if (isSigningPoolRegistrationAsOwner) {
+      // input should not be given with a path
+      // the path is not used, but we check just to avoid potential confusion of developers using this
+      Precondition.check(!input.path, "stake pool registration: inputs should not contain the witness path");
+    }
+  }
+
+  Precondition.checkIsArray(inputs, "outputs not an array");
```
refi93, line 126: the error message does not match the data being checked

src/cardano.js
```diff
@@ -1,49 +1,198 @@
...
+  for (const output of outputs) {
+    Precondition.checkIsValidAmount(output.amountStr, "invalid amount in output");
+    if (output.addressHex) {
+      Precondition.checkIsHexString(output.addressHex, "invalid address in output");
+      Precondition.check(output.addressHex.length <= 128 * 2, "invalid address in output: too long");
+    } else if (output.spendingPath) {
```
refi93, line 132: we probably shouldn't allow outputs given as paths for stakepool registrations as it may be abused to extract addresses from the transaction by the counterparty

src/cardano.js
```diff
@@ -62,22 +211,22 @@ export function serializeAddressInfo(
...
-    Precondition.check(stakingKeyHash.length == 28); // TODO some global constant for key hash length
+    Precondition.check(stakingKeyHash.length === KEY_HASH_LENGTH, "invalid staking key hash length");
     stakingInfoBuf = stakingKeyHash;
   } else if (!stakingPath && !stakingKeyHashHex &&  stakingBlockchainPointer) {
     stakingChoice = stakingChoices.BLOCKCHAIN_POINTER;
     stakingInfoBuf = Buffer.alloc(3 * 4); // 3 x uint32
-    
-    Precondition.checkIsUint32(stakingBlockchainPointer.blockIndex);
+
+    Precondition.checkIsUint32(stakingBlockchainPointer.blockIndex, "invalid blockchain pointer");
```
refi93, line 225: I would move the error message strings out to an object and reference the keys instead - it would allow for convenient reusing and less cluttered validation logic

src/Ada.js
```diff
@@ -527,65 +588,136 @@ export default class Ada {
...
+        for (const relay of poolParams.relays) {
+          const response = await _send(
+            P1_STAGE_CERTIFICATES,
+            APDU_INSTRUCTIONS.RELAYS,
+            cardano.serializePoolRelayParams(relay)
+          );
+          Assert.assert(response.length === 0);
+        }
+
+        const mdResponse = await _send(
```
refi93, line 707: mdResponse is imho too cryptic of a name, I would name it metadataResponse

src/Ada.js
```diff
@@ -527,65 +588,136 @@ export default class Ada {
...
+          P1_STAGE_CERTIFICATES,
+          APDU_INSTRUCTIONS.METADATA,
+          cardano.serializePoolMetadataParams(poolParams.metadata)
+        );
+        Assert.assert(mdResponse.length === 0);
```
refi93, line 712: I see those assertions being introduced with this feature only, do we really need/want them and if so, it would make sense to wrap the _send function into an "expectEmptyResponse" helper function instead of asserting explicitly all the time.

But imho asserting on ledger responses in js is rather redundant given we already do that on the hw app layer and such situations should not happen in the first place provided ledger firmware/js is implemented properly with proper input validation.

src/cardano.js
```diff
@@ -92,11 +241,216 @@ export function serializeAddressInfo(
...
+    ipv4Buf = Buffer.concat([yesBuf, ipBytes]);
+  } else {
+    ipv4Buf = noBuf;
+  }
+
+  let ipv6Buf: Buffer;
+  if (params.ipv6) {
+    const errorMsg = "invalid ipv6";
+    Precondition.checkIsString(params.ipv6, errorMsg);
+    let ipParts = params.ipv6.split(':');
```
refi93, line 367: this can be condensed into 

```
ipv6Address.split(':').join('');
```

refi93 (changes requested):
see comments

src/Ada.js
janmazak: I've refactored the assertions. For each `_send` in tx signing, the length of the response is now asserted (not only the empty ones). It's not really needed, but checking buffer lengths is cheap, so I don't feel any urge to remove it altogether.

refi93: it clutters the code IMO, making it harder to read for future contributors. I agree there is no harm in the additional check, and we can retain it, but I would suggest introducing a _sendAndExpectEmptyResponse function which would encapsulate this behavior then (feel free to find some more suitable name)

refi93: sorry I see now that the repetitive assertions on response length dissapeared, so disregard my previous comment

src/Ada.js
```diff
@@ -482,10 +558,16 @@ export default class Ada {
     const P1_STAGE_WITNESSES = 0x0a;
     const P2_UNUSED = 0x00;
 
-    const _send = (p1, p2, data) =>
-      this.send(CLA, INS.SIGN_TX, p1, p2, data).then(
-        utils.stripRetcodeFromResponse
+    const self = this;
+    const _send = async function(p1, p2, data, expectedResponseLength = 0) {
```
refi93, line 562: I would make expectedResponseLength undefined by default and disable the assertion for the response length if its undefined - this way the call can be used even for variable length responses and it's imho not very intuitive to do such check by default, better have the developer explicitly enable it by making them pass the expected length explicitly

janmazak: That's how I tried to write it at first, but for this particular case, all the responses have fixed length known in advance. So for a generic solution I would use the suggested way, but for signTransaction, I don't see much point in adding zeros: only two places use non-zero value and about a dozen or more use 0 for the length of the response.

PR #14, opened 2020-11-25
Allow single key export for older app versions

refi93:
Motivation: With the introduction of bulk key export in the latest app version we also introduced an app version check that prevents older Ledger apps from working with the new Ledgerjs even if the integrating party is still using single key export which may represent additional hassle for them (e.g. communicating to customers a breaking change requiring an arguably unnecessary update of their app)

Change: Drop the version check for single key export as the js code is actually backwards compatible with the older Ledger Cardano app

Testing: Tested with `yarn test-integration --grep getExtendedPublicKey` by setting a higher version number to check and verifying that single key export goes through whereas multiple key export fails. I also double-checked with Adalite which still uses the old ledgerjs (hence APDUs as well) that single key export works fine with the new Ledger app

### DIFF
--- a/src/Ada.js
+++ b/src/Ada.js
@@ -202,6 +202,10 @@ const DeviceErrorMessages = {
   [DeviceErrorCodes.ERR_UNSUPPORTED_ADDRESS_TYPE]: "Unsupported address type"
 };
 
+const Errors = {
+  INCORRECT_APP_VERSION: "Operation not supported by the Ledger device, make sure to have the latest version of the Cardano app installed"
+}
+
 export const getErrorDescription = (statusCode: number) => {
   const statusCodeHex = `0x${statusCode.toString(16)}`;
   const defaultMsg = `General error ${statusCodeHex}. Please consult ${GH_ERRORS_LINK}`;
@@ -271,6 +275,9 @@ export default class Ada {
   }
 
   async _getVersion(): Promise<GetVersionResponse> {
+    // moving getVersion() logic to private function in order
+    // to disable concurrent execution protection done by this.transport.decorateAppAPIMethods()
+    // when invoked from within other calls to check app version
     const _send = (p1, p2, data) =>
       this.send(CLA, INS.GET_VERSION, p1, p2, data).then(
         utils.stripRetcodeFromResponse
@@ -308,22 +315,23 @@ export default class Ada {
     return this._getVersion();
   }
 
-  async _checkLedgerCardanoAppVersion(minMajor: number, minMinor: number) {
+  async _isLedgerAppVersionAtLeast(minMajor: number, minMinor: number): Promise<boolean> {
     const version = await this._getVersion();
     const major = parseInt(version.major);
     const minor = parseInt(version.minor);
-    const patch = parseInt(version.patch);
 
-    const msg = "Operation not supported by the Ledger device, make sure to have the latest version of the Cardano app installed";
+    if (isNaN(major) || isNaN(minor))
+      return false;
 
-    if (isNaN(major) || isNaN(minor) || isNaN(patch))
-      throw new Error(msg);
+    return (major >= minMajor) && (major>minMajor || minor>=minMinor);
+  }
 
-    if (major < minMajor)
-      throw new Error(msg);
+  async _ensureLedgerAppVersionAtLeast(minMajor: number, minMinor: number) {
+    const versionCheckSuccess = await this._isLedgerAppVersionAtLeast(minMajor, minMinor)
 
-    if ((major === minMajor) && (minor < minMinor))
-      throw new Error(msg);
+    if (!versionCheckSuccess) {
+      throw new Error(Errors.INCORRECT_APP_VERSION);
+    }
  }
 
   /**
@@ -337,7 +345,7 @@ export default class Ada {
    *
    */
   async getSerial(): Promise<GetSerialResponse> {
-    await this._checkLedgerCardanoAppVersion(1, 2);
+    await this._ensureLedgerAppVersionAtLeast(1, 2);
 
     const _send = (p1, p2, data) =>
       this.send(CLA, INS.GET_SERIAL, p1, p2, data).then(
@@ -386,7 +394,7 @@ export default class Ada {
     }
 
     if (paths.length > 1) {
-      await this._checkLedgerCardanoAppVersion(2, 1);
+      await this._ensureLedgerAppVersionAtLeast(2, 1);
     }
 
     const _send = (p1, p2, data) =>
@@ -406,15 +414,15 @@ export default class Ada {
       let response: Buffer;
       if (i === 0) {
         // initial APDU
-        const remainingKeysData = utils.uint32_to_buf(paths.length - 1);
+
+        // passing empty Buffer for backwards compatibility
+        // of single key export on Ledger app version 2.0.4
+        const remainingKeysData = paths.length > 1
+          ? utils.uint32_to_buf(paths.length - 1)
+          : Buffer.from([]);
 
         response = await wrapRetryStillInCall(_send)(
-          P1_INIT, P2_UNUSED,
-          // remainingKeysData cannot be passed for single key export
-          // to maintain backwards compatibility with Ledger app version 2.0.4
-          paths.length > 1
-          ? Buffer.concat([pathData, remainingKeysData])
-          : pathData
+          P1_INIT, P2_UNUSED, Buffer.concat([pathData, remainingKeysData])
         );
       } else {
         // next key APDU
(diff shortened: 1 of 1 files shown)


janmazak (approved):

refi93:
Added a patch to allow backwards compatibility with ledger app 2.0.4 (i.e. the currently released version) of the signTransaction() function as the isSigningPoolRegistrationAsOwner flag breaks backwards compatibility

src/Ada.js
```diff
@@ -610,11 +628,16 @@ export default class Ada {
           ? MetadataCodes.SIGN_TX_METADATA_YES
           : MetadataCodes.SIGN_TX_METADATA_NO
         ),
-        utils.uint8_to_buf(
-          isSigningPoolRegistrationAsOwner
-          ? PoolRegistrationCodes.SIGN_TX_POOL_REGISTRATION_YES
-          : PoolRegistrationCodes.SIGN_TX_POOL_REGISTRATION_NO
-        ),
+        (() => {
+          if (appHasStakePoolOwnerSupport) {
```
refi93, line 632: this is relatively ugly and I think it would be prone to weird bugs in the mid-to-long term, so I would personally prefer to remove this workaround based on app version as soon as app version 2.1.0 is launched

src/Ada.js
```diff
@@ -409,7 +421,11 @@ export default class Ada {
 
         response = await wrapRetryStillInCall(_send)(
           P1_INIT, P2_UNUSED,
-          Buffer.concat([pathData, remainingKeysData])
+          // remainingKeysData cannot be passed for single key export
+          // to maintain backwards compatibility with Ledger app version 2.0.4
+          paths.length > 1
+          ? Buffer.concat([pathData, remainingKeysData])
+          : pathData
```
janmazak, line 428: I would assign this to a variable outside of the function call.

src/Ada.js
```diff
@@ -566,7 +582,9 @@ export default class Ada {
     );
 
     if (isSigningPoolRegistrationAsOwner)
-      await this._checkLedgerCardanoAppVersion(2, 1);
+      await this._checkLedgerCardanoAppVersionOrThrow(2, 1);
+
+    const appHasStakePoolOwnerSupport = await this._checkLedgerCardanoAppVersion(2, 1)
```
janmazak, line 587: We can first compute the value of this variable and then throw
if (isSigningPoolRegistrationAsOwner && !appHasStakePoolOwnerSupport)

src/Ada.js
janmazak: I agree it is only good as a temporary patch and we want to remove it as soon as possible.

src/Ada.js
refi93: done

janmazak: Just the semicolon is missing.

src/cardano.js
```diff
@@ -91,6 +91,10 @@ export const TxErrors = {
   METADATA_INVALID: "invalid metadata",
 }
 
+export const Errors = {
```
janmazak, line 94: I feel unsure about having this type of error in cardano.js which tries to put together cardano-related stuff. Errors related to the Ledger device are in Ada.js (see DeviceErrorMessages).

refi93: I see, DeviceErrorMessages are mapped to device error codes though, but makes sense to move the structure to Ada.js

src/Ada.js
```diff
@@ -308,22 +308,30 @@ export default class Ada {
     return this._getVersion();
   }
 
-  async _checkLedgerCardanoAppVersion(minMajor: number, minMinor: number) {
+  async _checkLedgerCardanoAppVersion(minMajor: number, minMinor: number): Promise<boolean> {
```
ppershing, line 311: why do we keep the name? Whould this be `_isLedgerAppVersionCompatible` or something like that

src/Ada.js
```diff
@@ -379,15 +387,16 @@ export default class Ada {
...
-    await this._checkLedgerCardanoAppVersion(2, 1);
-
     // validate the input
     Precondition.checkIsArray(paths);
     for (const path of paths) {
       Precondition.checkIsValidPath(path);
     }
 
+    if (paths.length > 1) {
+      await this._checkLedgerCardanoAppVersionOrThrow(2, 1);
```
ppershing, line 397: Now I am confused -- above is major=1, minor=2, here it is opposite. What should the semantic of the call?

refi93: are you sure? I see minMajor=2 and minMinor=1 But I agree the "min" prefix might be confusing

refi93: I see, sorry

refi93: indeed, getSerial call seems to have been introduced in Ledger app version 1.2 based on my search in older code: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/527e5b0fa0ab31b37395808248cef01875ed0eae/src/Ada.js#L316 so It's just a coincidence and the version check for getSerial call is actually supposed to check for version 1.2. But given that we version 1 is Byron-era cardano app which would no longer work anyway, it's arguably a redundant check

src/Ada.js
refi93: fair point - as the function returns a boolean instead of throwing, your suggested name makes more sense

src/Ada.js
```diff
@@ -308,22 +312,30 @@ export default class Ada {
     return this._getVersion();
   }
 
-  async _checkLedgerCardanoAppVersion(minMajor: number, minMinor: number) {
+  async _isLedgerAppVersionCompatible(minMajor: number, minMinor: number): Promise<boolean> {
     const version = await this._getVersion();
     const major = parseInt(version.major);
```
ppershing, line 317: why isn't this part of getVersion? :thinking:

src/Ada.js
```diff
@@ -308,22 +312,30 @@ export default class Ada {
...
     const minor = parseInt(version.minor);
     const patch = parseInt(version.patch);
 
-    const msg = "Operation not supported by the Ledger device, make sure to have the latest version of the Cardano app installed";
-
     if (isNaN(major) || isNaN(minor) || isNaN(patch))
-      throw new Error(msg);
+      return false
 
     if (major < minMajor)
```
ppershing, line 324: this whole logic is very obscure. It would read way better if you had `(major >= minMajor && (major>minMajor || minor>=minMinor)`. However, I am still not sold on this concept -- this assumes full future compatibility which is weird. In fact, change in majorVersion should be breaking change imho

src/Ada.js
```diff
@@ -405,11 +418,18 @@ export default class Ada {
       let response: Buffer;
       if (i === 0) {
         // initial APDU
-        const remainingKeysData = utils.uint32_to_buf(paths.length - 1);
+        let payload;
+        if (paths.length > 1) {
```
ppershing, line 422: I hate this workaround. Why don't we pass it to getSinglePublicKey call?

src/Ada.js
refi93: if you mean the parseInt() I would guess it accounts for letters in the version name, but then this code would break so I agree that this is overly cumbersome and it makes sense to just return numbers

src/Ada.js
refi93: I agree the name `_isLedgerAppVersionCompatible` "lures" the reader into assuming the result of the call guarantees compatibility and possibly does some sophisticated app feature checks which is not true. Imho considering a major version number increase to be breaking on all calls is a bit too restrictive, as it may be breaking just for a part of the calls.

Therefore I think a name like `_isLedgerAppVersionAtLeast` would be more suitable and actually matching up with what the function does. And I would try not to deal with forward compatibility or any more sophisticated way to determine compatibility for now. In fact, as an ordinary user you cannot really do much if your Ledger has a higher app version than your software wallet supports so even if we returned some kind of comprehensive error message regarding forward compatibility, it would be a dead avenue for the users as you cannot really roll back with Ledger Live as far as I know.

src/Ada.js
refi93: getExtendedPublicKey is implemented as a special case of its bulk version, so that wouldn't really work, but I can see having it rewritten in a less awkward way:

```
          const remainingKeysData = paths.length > 1
          ? utils.uint32_to_buf(paths.length - 1)
          : Buffer.from([]);

        response = await wrapRetryStillInCall(_send)(
          P1_INIT, P2_UNUSED, Buffer.concat([pathData, remainingKeysData])
        );
```

src/Ada.js
refi93: or I would rather leave getVersion as-is for now, this does not really seem like a compelling reason to change a call that others might depend on e.g. for additional version checks that would break due to type mismatches or something alike

src/Ada.js
```diff
@@ -294,36 +308,23 @@ export default class Ada {
...
     const minor = parseInt(version.minor);
-    const patch = parseInt(version.patch);
 
-    const msg = "Operation not supported by the Ledger device, make sure to have the latest version of the Cardano app installed";
+    if (isNaN(major) || isNaN(minor))
+      return false;
 
-    if (isNaN(major) || isNaN(minor) || isNaN(patch))
-      throw new Error(msg);
+    return major >= minMajor && (major>minMajor || minor>=minMinor);
```
janmazak, line 319: I'd keep spaces around operators, to be consistent.

janmazak (approved):

PR #18, opened 2020-12-02
Fix flow errors, introduce invariant() helper function

refi93:
Motivation: In the implementation of pool owner support 16 flow errors were apparently introduced. I was trying to find a convenient way to do type narrowing with our Asser.assert() utility function but Flow does not seem to support custom type narrowing functions.

What I found though is that Flow has an apparently hard-coded behavior that whatever statement is in a function called "invariant" it results in corresponding type narrowing, so this PR is leveraging that property which is apparently present in flow for over 4 years: https://github.com/facebook/flow/issues/2617 and seems to be recognized by the community, though it is regarded as obscure: https://github.com/facebook/flow/issues/6052

I was also considering whether to completely replace `assert.Assert()` with `invariant()` but most of the instances were
 run time validation checks, like response length where I found it a bit weird to be enforcing an "invariant" as I perceive that as something that is "provably" true based on the logic of the code itself rather than enforcement of some external factor

Changes:
* introduce custom `invariant()` function that does the same as `Assert.assert()` but is meant to do type narrowing in Flow, leveraging the fact that Flow recognizes it as such

Testing:
* `yarn flow` passes with 0 errors. I also tried putting the negations of the conditions I checked with `invariant` and they actually made `yarn flow` fail with corresponding errors, confirming that flow is not completely silenced by casting the checked values to `any` or something similar

### DIFF
--- a/src/Ada.js
+++ b/src/Ada.js
@@ -19,7 +19,7 @@
 import type Transport from "@ledgerhq/hw-transport";
 import { TransportStatusError } from "@ledgerhq/hw-transport";
 
-import utils, { Precondition, Assert } from "./utils";
+import utils, { Precondition, Assert, invariant } from "./utils";
 import cardano, { CertificateTypes, AddressTypeNibbles, TxErrors } from "./cardano";
 
 const CLA = 0xd7;
@@ -737,6 +737,8 @@ export default class Ada {
       // we are done for every certificate except pool registration
 
       if (type === CertificateTypes.STAKE_POOL_REGISTRATION) {
+        invariant(poolParams != null);
+
         // additional data for pool certificate
         const APDU_INSTRUCTIONS = {
           POOL_PARAMS: 0x30,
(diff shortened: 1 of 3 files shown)


src/Ada.js
```diff
@@ -744,6 +744,8 @@ export default class Ada {
       // we are done for every certificate except pool registration
 
       if (type === CertificateTypes.STAKE_POOL_REGISTRATION) {
+        invariant(poolParams != null)
```
janmazak, line 747: Semicolon, also below.

janmazak (commented):
I have explored some other options, mainly predicate functions (https://flow.org/en/docs/types/functions/#toc-predicate-functions), but this seems simplest, and likely to work for many years to come... 

But I still see 8 errors when running `yarn flow`.

refi93:
@janmazak fixed the semicolons, and you also need to run `yarn build` in order for the changes to be reflected when running `yarn flow`

refi93:
ok I realized why after running `yarn build` the flow errors disappeared - the `yarn flow` command is checking also the `lib` folder where the built js files reside, so it basically checks each file twice - once in `src` and the seconf time in `lib` folder :facepalm: So I'm narrowing down the flow command to scan just `src` folder

src/Ada.js
```diff
@@ -939,6 +942,7 @@ export default class Ada {
     // witnesses
     const witnesses = [];
     for (const path of witnessPaths) {
+      invariant(path != null);
```
janmazak, line 945: The witness path is not explicitly validated (neither here nor above where witnessPaths is constructed).
(Maybe it is checked elsewhere, for owners, inputs etc., but calling checkIsValidPath would not hurt.)

janmazak (approved):

src/Ada.js
refi93: this is actually done in validateTransaction and signTx_getWitness does the validation of the path being passed as well with a comprehensive error message, so IMHO an additional validation may clutter the code more than making it more readable, as the reader may even wonder if the path can be invalid at that point

refi93: ok, adding the check to the construction of witnessPaths

refi93: finally, I added the path check here as adding it in witnessPaths construction triggered flow errors related to the paths checked potentially not being arrays, cluttering the code with more invariant checks

PR #20, opened 2020-12-07
Develop rebased release 2.1.0

refi93:
Motivation: the "develop" branch of this repo got missaligned with cardano-foundation's "shelley" branch (because of subsequent rebases made in this repo), resulting in merge conflicts when trying to promote the changes released from this repo. We can either create a merge commit and introduce several tens of repeated commits or just rebase the develop of this repo to cardano-foundations's "shelley" branch

This is a rebased version of develop branch which we can overwrite develop with in order to align this repo's develop with cardano-foundation's "shelley" branch: https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/tree/shelley which should allow merging the changes smoothly into the upstream repo.

A side-effect of this force-push to develop would be that the already existing PRs would get out of sync but the conflicts should be easily solvable by mass-dropping the dangling old develop commits when rebasing. The existing github (pre)releases are bound to tags which will persist even after the force-push

Changes:
* should be the same as current develop with the addition of flow types from https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/commit/53731c08f384af7c226a37ac2a50f0d6b4f45b0d with the difference that CertTypes type has been unified with the CertificateTypes (something we should notify Yoroi about, as they seem to be using CertTypes: https://github.com/Emurgo/yoroi-frontend/search?q=CertTypes)

Testing:
* changes verifiable locally by`git checkout develop-rebased-release-2.1.0 && git diff develop` which show only the changes introduced in https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/commit/53731c08f384af7c226a37ac2a50f0d6b4f45b0d + the type refactor mentioned above which is expected
* `yarn flow` passes
* `yarn test-integration` passes

DO NOT MERGE, instead this is meant to be force-pushed into develop, once approved

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cardano-foundation/ledgerjs-hw-app-cardano",
-  "version": "2.0.2-rc.1",
+  "version": "2.1.0",
   "files": [
     "lib"
   ],


janmazak (approved):
It looks reasonable to me. I have not tried to run it with various versions of the Ledger app, but I don't see any problems with the code.

PR #15, opened 2020-12-01
Operational certificate signing and cold key derivation

refi93:
Motivation: Add support operational certificate signing and cold key derivation inspired by existing signTransaction and getExtendedPublicKey calls, as part of the new operator app feature

depends on https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/48

Changes:
* introduce getExtendedColdPublicKey() call
* introduce signOperationalCertificate() call
* refactor custom `_send()` function construction to a common `buildSendFn` builder function
* add corresponding integration tests
* fix `GetVersionResponse` type to match reality (version numbers were actually numbers, not strings)

Testing:
```
yarn test-integration --grep getExtendedPoolColdPublicKey
yarn test-integration --grep signOperationalCertificate
```

TODOs:
* restrict/skip tests if ledger app is running with operator features enabled (probably skipping based on exception thrown)
* add docs

PR #24, opened 2021-01-19
pool operator app

janmazak:
Tests for the new use case (signing tx as operator) are missing.


janmazak:
TODO:
(JM) doplnenie testov pre novy usecase
(JM) rebase nad operational-certificate-support (after #15 is finished)
(RK) review (az po rebase?)

refi93:
TODO: port changes from https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/15

src/Ada.js
```diff
@@ -81,6 +83,11 @@ export type StakingBlockchainPointer = {|
   certificateIndex: number
 |}
 
+export type PoolKeyParams = {|
+  keyHashHex: string,
```
refi93, line 87: shouldn't both be optional and enforce at runtime that either the former or latter is present?

src/Ada.js
```diff
@@ -570,6 +582,32 @@ export default class Ada {
     await _send(P1_DISPLAY, P2_UNUSED, data, 0);
   }
 
+  async getExtendedPoolColdPublicKey(
+    path: BIP32Path
+  ): Promise<GetExtendedPublicKeyResponse> {
+    await this._ensureLedgerAppVersionAtLeast(2, 1); // TODO update version number
```
refi93, line 588: this can be updated to 2.2 at least, if not 2.3 (foreseen pool operator app support release)

src/cardano.js
```diff
@@ -508,10 +749,10 @@ export function serializePoolRelayParams(
   const RELAY_YES = 2;
 
   const yesBuf = Buffer.alloc(1);
-  yesBuf.writeUInt8(RELAY_YES);
+  yesBuf.writeUInt8(RELAY_YES, 0);
```
refi93, line 752: it's clearer that we are actually writing the byte from the beginning of the buffer and not after the preallocated chunk, but I'd rather rewrite that to `yesBuf = Buffer.concat([utils.uint8_to_buf(RELAY_YES)])` rather than dealing with offsets which a casual reader needs to look up in nodejs docs

refi93: anyway, a similar clarity issue exist below with ipv4 and ipv6 serialization and given the ongoing refactor it's questionable whether it's worth doing anything about it here if it works

src/cardano.js
```diff
@@ -564,32 +805,32 @@ export function serializePoolRelayParams(
...
-  typeBuf.writeUInt8(type);
+  typeBuf.writeUInt8(type, 0);
 
   switch(type) {
     case 0:
       return Buffer.concat([typeBuf, portBuf, ipv4Buf, ipv6Buf]);
 
     case 1:
-      Precondition.check(dnsBuf, TxErrors.CERTIFICATE_POOL_RELAY_MISSING_DNS);
+      Precondition.check(!!dnsBuf, TxErrors.CERTIFICATE_POOL_RELAY_MISSING_DNS);
```
refi93, line 815: I'd rather check explicitly for dnsBuf != null as generally the "truthiness" of values in JS induces headaches when reasoning about the code

src/cardano.js
janmazak: I added the offsets just because flow said they must be there :) I've changed a bit to use uint8_to_buf and we will deal with it after the ts refactor.

src/cardano.js
```diff
@@ -448,57 +695,55 @@ export function serializePoolInitialParams(
...
     utils.uint64_to_buf(params.margin.denominatorStr),
-    utils.hex_to_buf(params.rewardAccountHex),
-    utils.uint32_to_buf(params.poolOwners.length),
-    utils.uint32_to_buf(params.relays.length)
   ]);
 }
 
-export function serializePoolOwnerParams(
-  params: PoolOwnerParams
+export function serializePoolParamsRewardAccount(
```
refi93, line 706: I think we are missing here the function to serialize the reward account given as a staking path

test/src/integration/signTxPoolRegistration.js
```diff
@@ -432,3 +468,72 @@ describe("signTxPoolRegistrationReject", async () => {
...
+      sampleFeeStr,
+      sampleTtlStr,
+      [cert],
+      [],
+      null
+    );
+    expect(response).to.deep.equal(results.poolRegistrationOperatorNoOwnersNoRelays);
+  });
+
+  it("Should correctly witness pool registration with one owner-operator as operator", async () => {
```
refi93, line 504: passing owners as paths when signing a pool registration certificate as an operator is currently not allowed AFAIK and looking at the testcase it doesn't seem the case, so the description seems misleading

janmazak:
With backwards compatibility added, this is the final pre-TypeScript version of this library.

CHANGELOG.md
```diff
@@ -0,0 +1,76 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+ 
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
```
refi93, line 6: please update change log accordingly

janmazak:
Replaced by operatorapp-ts.

PR #27, opened 2021-01-28
Multiassets

janmazak:
Not 100% finished and needs some polishing.


src/Ada.js
```diff
@@ -44,21 +44,37 @@ export type InputTypeUTxO = {|
   path: ?BIP32Path
 |};
 
-export type OutputTypeAddress = {|
+
+export type TokenAmount =
```
refi93, line 48: in Trezor we ended up renaming the the entities to remain closer to the developer docs terminology (https://developers.cardano.org/en/development-environments/native-tokens/native-tokens-glossary/), now it's:

TokenGroup -> AssetGroup
TokenAmount -> Token
tokenAmounts -> tokens

Trezor protobuf definitions for reference: https://github.com/trezor/trezor-firmware/blob/master/common/protob/messages-cardano.proto#L149

src/Ada.js
```diff
@@ -634,10 +655,20 @@ export default class Ada {
       const data = Buffer.concat([
         utils.uint8_to_buf(networkId),
         utils.uint32_to_buf(protocolMagic),
+        utils.uint8_to_buf(
```
refi93, line 658: If I understand correctly this is a breaking change, so don't forget to check for the proper version of the Ledger app. We can also add some backward-compatibility hacks, but this is probably a good opportunity to do a clean cut

src/Ada.js
```diff
@@ -891,26 +931,13 @@ export default class Ada {
...
-        // validation should catch invalid outputs
-        Assert.assert(false, "invalid output type");
-      }
+      await signTx_addOutput(output);
     }
 
     await signTx_setFee(feeStr);
 
-    await signTx_setTtl(ttlStr);
+    if (ttlStr)
```
refi93, line 939: I'd wrap it in curly braces for the sake of maintainability/consistency, same for validityIntervalStartStr

src/cardano.js
```diff
@@ -241,6 +263,13 @@ export function validateTransaction(
     Precondition.checkIsHexString(metadataHashHex, TxErrors.METADATA_INVALID);
     Precondition.check(metadataHashHex.length == 32 * 2, TxErrors.METADATA_INVALID);
   }
+
+  //  validity interval start
+  if ((validityIntervalStartStr !== null) && (validityIntervalStartStr !== undefined)) {
```
refi93, line 268: `validityIntervalStartStr != null` is a more concise way to express the same

https://stackoverflow.com/questions/2559318/how-to-check-for-an-undefined-or-null-variable-in-javascript

src/utils.js
```diff
@@ -243,12 +254,7 @@ export function buf_to_amount(data: Buffer): string {
 export function amount_to_buf(amount: string): Buffer {
```
refi93, line 254: given that we are introducing token amounts which are not subject to the same constraints as ada amounts it imho makes sense to rename this to `ada_amount_to_buf` - or just drop the check for max lovelace supply for the sake of simplicity - TBH I don't see that much added value in the check as that's the responsibility of the blockchain and may not even be true for some testnets

refi93 (changes requested):
see comments

src/utils.js
janmazak: Well, it only reflects the check that Ledger is doing. We might consider dropping it there. 
(I think the check is there partially for added protection against attacks --- not much, of course, but every assert and validation makes the life of an attacker trying to corrupt the memory harder.)

refi93: makes sens, in the end it's a really niche corner case, so I'm ok with keeping the constraint as well though to not mix it accidentally with token amounts I'd rename it at least

src/Ada.js
```diff
@@ -580,9 +591,34 @@ export default class Ada {
 
     const appHasStakePoolOwnerSupport = await this._isLedgerAppVersionAtLeast(2, 1);
     if (isSigningPoolRegistrationAsOwner && !appHasStakePoolOwnerSupport) {
-      throw Error(Errors.INCORRECT_APP_VERSION)
+      throw Error(Errors.INCORRECT_APP_VERSION);
+    }
+
+    // TODO replace this with a better mechanism for detecting ledger app capabilities
```
refi93, line 597: I would not be afraid to centralize that in a single mapping/helper function - I did something like that in trezor-connect recently and I quite liked it, it wasn't even that much work: https://github.com/trezor/connect/commit/fe4a55cac2e7ffe1bf6b8d5af55ad83f767b6460

src/Ada.js
```diff
@@ -788,43 +867,61 @@ export default class Ada {
...
     const signTx_awaitConfirm = async (): Promise<{
       txHashHex: string
     }> => {
+
+      // TODO remove after ledger app 2.2 is rolled out
+      let confirmP1;
+      if (appHasMultiassetSupport) {
+        confirmP1 = P1_STAGE_CONFIRM;
+      } else {
+        confirmP1 = 0x09; // needed for backward compatibility
```
refi93, line 920: I'd name that constant P1_STAGE_CONFIRM_LEGACY - even if it's "backwards compatibility" code better keep basic readability

src/cardano.js
```diff
@@ -314,6 +349,91 @@ export function serializeAddressParams(
...
+    throw new Error(TxErrors.OUTPUT_UNKNOWN_TYPE);
+  }
+
+  const numassetGroups =
+      (output.tokenBundle) ?
+      output.tokenBundle.length :
+      0;
+
+  return Buffer.concat([
+    utils.uint8_to_buf(outputType),
```
refi93, line 394: what's the reason behind reordering the parameters in this APDU respective to the old version? Wouldn't be the backwards-compatibility logic be less ugly, not requiring the splitting into two separate functions?

refi93 (changes requested):
see comments

refi93:
tested with Ledger fw 2.0.4 (a0bf957d1a919f64427dc13680e350cbe3945fdc) and 2.1.0 (d031a6d18e988537e90792fd40a13294528ca40f)  and works as expected (i.e. only features not supported in those releases fail, the rest passes)

src/cardano.js
janmazak: I find the order "ada amount, address, tokenAmounts" rather unintuitive and prefer address, ada amount, token amounts. The temporary need for backwards compatibility does not change this preference.

src/utils.js
```diff
@@ -149,6 +149,17 @@ export function buf_to_uint32(data: Buffer): number {
   return data.readUIntBE(0, 4);
 }
 
+export function uint64_to_buf(value: string): Buffer {
+  const i = safe_parseInt(value);
+  Precondition.checkIsUint64(i, "invalid uint64 value");
```
PeterBenc, line 154: One of the `cardano-hw-cli` tests fails because it has ADA amount bigger than an integer.

src/Ada.js
janmazak: I don't have a clear picture of this in my mind, so I would rather leave it as it is for this release, and tackle it in the subsequent refactoring.

src/utils.js
refi93: will be fixed by merging #29

PR #28, opened 2021-01-31
Refactors

ppershing:


src/Ada.js
```diff
@@ -425,8 +455,16 @@ export default class Ada {
...
    */
+
   async getExtendedPublicKeys(
     paths: Array<BIP32Path>
+  ): Promise<Array<GetExtendedPublicKeyResponse>> {
+    return this._getExtendedPublicKeys(this._send, paths);
+  }
+
+  async _getExtendedPublicKeys(
+    _send: SendFn,
```
refi93, line 466: what benefit do you see in injecting the "_send" function from outside if is fed to the method from inside the class anyway? Is this a preparation to extract the individual calls into pure functions?

refi93 (approved):
one clarifying question, otherwise LGTM

src/Ada.js
ppershing: Exactly. These functions are going to go out of the class, into own files, and finally we will get rid of _send as well but the exact way is a secret for now :-)

PR #29, opened 2021-02-05
Fix validation of numerical parameters passed as strings

refi93:
Motivation: big numbers validation in various places was incorrect because it was casting the value to native js numbers which do not work well over a certain threshold lower than 2^64, The used safe_parseInt function thrown (as it was designed to) for numbers over max safe integer in javascript, so numbers over that threshold (~2^53) were basically causing ledgerjs validation to fail and the transaction was not passed to Ledger.

I was originally planning to just extend the original simplistic validation of integer strings but it got awkward when I wanted to generalize it for uint64 and ADA amount (adding a min-max range to the function interface) and there is also validation comparing pool margin numerator and denominator.

So it felt kind of shortsighted to create a relatively ugly helper function that would require further "hacking" as more numbers with different constraints would be introduced. Therefore I went with JSBI library implementing ES2020 BigInt. I considered using BigInt directly but it seems to be too "bleeding edge" and it may not work in older, but still relatively new browsers and developers may face unnecessary issues with that. Moreover, it seems tricky to be translated by Babel: https://javascript.info/bigint#polyfills and the plugin that I found for it https://www.npmjs.com/package/babel-plugin-transform-bigint doesn't seem very stable and well adopted and I didn't manage to get it working, it was failing with `Error: Cannot find module './jsbi.mjs'`

Changes:
* refactor validation of numbers to use BigInt (JSBI)
* add testcase containing a transaction with big numbers

Testing:
* `yarn test-integration` passes

### DIFF
--- a/src/cardano.js
+++ b/src/cardano.js
@@ -83,6 +83,7 @@ export const TxErrors = {
   CERTIFICATE_POOL_INVALID_PLEDGE: "invalid pledge in a pool registration certificate",
   CERTIFICATE_POOL_INVALID_COST: "invalid cost in a pool registration certificate",
   CERTIFICATE_POOL_INVALID_MARGIN: "invalid margin in a pool registration certificate",
+  CERTIFICATE_POOL_INVALID_MARGIN_DENOMINATOR: "pool margin denominator must be a value between 1 and 10^15",
   CERTIFICATE_POOL_INVALID_REWARD_ACCOUNT: "invalid reward account in a pool registration certificate",
   CERTIFICATE_POOL_OWNERS_NOT_ARRAY: "owners not an array in a pool registration certificate",
   CERTIFICATE_POOL_OWNERS_TOO_MANY: "too many owners in a pool registration certificate",
@@ -233,9 +234,7 @@ export function validateTransaction(
         for (const token of assetGroup.tokens) {
           Precondition.checkIsHexString(token.assetNameHex, TxErrors.OUTPUT_INVALID_ASSET_NAME);
           Precondition.check(token.assetNameHex.length <= TOKEN_NAME_LENGTH * 2, TxErrors.OUTPUT_INVALID_ASSET_NAME);
-
-          const amount = utils.safe_parseInt(token.amountStr);
-          Precondition.checkIsUint64(amount);
+          Precondition.checkIsUint64Str(token.amountStr);
         }
       }
     }
@@ -246,9 +245,7 @@ export function validateTransaction(
 
   //  ttl
   if (ttlStr != null) {
-    const ttl = utils.safe_parseInt(ttlStr);
-    Precondition.checkIsUint64(ttl, TxErrors.TTL_INVALID);
-    Precondition.check(ttl > 0, TxErrors.TTL_INVALID);
+    Precondition.checkIsPositiveUint64Str(ttlStr, TxErrors.TTL_INVALID);
   }
 
   // certificates
@@ -272,9 +269,7 @@ export function validateTransaction(
 
   //  validity interval start
   if (validityIntervalStartStr != null) {
-    const validityIntervalStart = utils.safe_parseInt(validityIntervalStartStr);
-    Precondition.checkIsUint64(validityIntervalStart, TxErrors.VALIDITY_INTERVAL_START_INVALID);
-    Precondition.check(validityIntervalStart > 0, TxErrors.VALIDITY_INTERVAL_START_INVALID);
+    Precondition.checkIsPositiveUint64Str(validityIntervalStartStr, TxErrors.VALIDITY_INTERVAL_START_INVALID);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -345,8 +349,6 @@ export default {
 
   str_to_path,
 
-  safe_parseInt,
```
refi93, line 348: removed duplicate export, check line 359

ppershing:
I am not sure I like adding another external dependency, just for the sake of parsing few numbers. Are we sure we
1) need to allow transactions with such large values? (though I agree it would be nice)
2) can't we use native bigint? https://caniuse.com/bigint seems to suggest that front-runners (edge, Firefox, Chrome are very happy with it. Unless IOHK uses ages-old electron (i.e. node <v10), we should be fine with native approach). The only struggle I see is Safari which added the support only recently (Sept 2020). Alternatively, I know "bignumber.js" to be quite reliable, has 10x the npm downloads and shouldn't need transpilation

ppershing:
I suggest we open discussions with Emurgo/IOHK whether we can actually use *native* bigints in the followup refactor. It would be nice to get rid of all the "str" values and use native ones. (We could still throw LossOfPrecisionError if we detect using float for values >2**53)

refi93:
@ppershing regarding point 1 - there can be legitimate transactions on testnet that big (we got a bug report from IOHK related to that) and the custom tokens also let anybody deal with such big numbers even on mainnet

regarding point 2 - using native bigint in a library potentially used by anybody is imho a needless risk towards the products that are integrating with it. Would they be able to do something if they for example needed to work with an older Safari browser or node version 8 (the latest LTS)? And even if we confirm with IOHK/Emurgo they are OK with BigInt, the library is public, so there may still be somebody else getting blocked by this we didn't account for. bignumber.js is another alternative of course, though it also offers operations with decimal numbers which we don't really need and doesn't have the benefit of being transpilable into native BigNumber once we choose t, though that's probably not that big of a deal

Finally, we may need to work pretty soon with negative numbers (minting of tokens) as well so that can complicate things further if we wanted to keep our own implementation for parsing/validating numbers but that it's a fair point that line of thought falls into premature optimization.

All in all, to not hamper the upcoming refactoring and incur the cost of an additional library, I'd rewrite it to use our own function. Imho using native BigInt in the interface is risky even if we got the heads up from Emurgo/IOHK, ironically Adalite, our own product. may face the most issues as a browser based wallet meant not just for the most mainstream ones and there may not be a reasonable way to work around the problem should users with older browsers face issues with this. But that's a discussion we don't need to decide on now

refi93:
@ppershing I removed the external dependency, it turned out quite acceptable IMHO, you can check again

src/utils.js
```diff
@@ -150,8 +152,7 @@ export function buf_to_uint32(data: Buffer): number {
 }
 
 export function uint64_to_buf(value: string): Buffer {
-  const i = safe_parseInt(value);
-  Precondition.checkIsUint64(i, "invalid uint64 value");
+  Precondition.checkIsUint64Str(value, "invalid uint64 value");
 
   const data = bs10.decode(value);
```
ppershing, line 157: what? This really works? Do we have test for this? :thinking:

refi93: It's the original way of encoding amounts: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blame/develop/src/utils.js#L249 this function is just the renamed version of it with the interface generalized to deal with an arbitrary uint64 values

We don't test it directly, only via the integration tests that trigger APDU serialization relying on this. Indeed, it would be nice to set up unit tests which we currently don't have to transparently test such small but important pieces of logic

src/cardano.js
```diff
@@ -246,9 +244,8 @@ export function validateTransaction(
 
   //  ttl
   if (ttlStr != null) {
-    const ttl = utils.safe_parseInt(ttlStr);
-    Precondition.checkIsUint64(ttl, TxErrors.TTL_INVALID);
-    Precondition.check(ttl > 0, TxErrors.TTL_INVALID);
+    Precondition.checkIsUint64Str(ttlStr, TxErrors.TTL_INVALID);
+    Precondition.check(ttlStr !== "0", TxErrors.TTL_INVALID);
```
janmazak, line 248: I would suggest to wrap this into a precondition function because it looks suspicious and cannot be fully understood unless one is familiar with checkIsUint64Str in detail (e.g. what happens for "0x0").

src/cardano.js
```diff
@@ -446,13 +442,13 @@ export function serializePoolInitialParams(
...
-  const marginDenominator = utils.safe_parseInt(params.margin.denominatorStr);
-  Precondition.checkIsUint64(marginDenominator, TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
-  Precondition.check(marginNumerator >= 0, TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
-  Precondition.check(marginDenominator > 0, TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
-  Precondition.check(marginNumerator <= marginDenominator, TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
+  const marginNumeratorStr = params.margin.numeratorStr;
+  const marginDenominatorStr = params.margin.denominatorStr;
+  Precondition.checkIsUint64Str(marginNumeratorStr, TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
+  Precondition.checkIsUint64Str(marginDenominatorStr, TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
+  Precondition.check(marginNumeratorStr !== "0", TxErrors.CERTIFICATE_POOL_INVALID_MARGIN);
```
janmazak, line 449: Is zero margin forbidden? It was not, until now.

janmazak: Another issue is that Ledger validates denominator against another bound:
#define MARGIN_DENOMINATOR_MAX 1000000000000000ul // 10^15
VALIDATE(p->marginDenominator <= MARGIN_DENOMINATOR_MAX, ERR_INVALID_DATA);
The bound is there because otherwise we could not calculate and display the margin (a percentage with 2 decimal places) using built-in arithmetics (we would need to write our own division algorithm) --- afaik this is what IOHK approved, so I don't think we want to change it.

janmazak: Violation of this additional bound should probably use its own error msg (so that the user is not confused and knows that he is supplying correct uint64, but not one suitable for ledger).

refi93: I see, yes, I confused the numerator and denominator here, will fix that

src/utils.js
```diff
@@ -15,8 +15,10 @@ const BECH32_ALPHABET =
 // We use bs10 as an easy way to parse/encode amount strings
 const bs10 = basex("0123456789");
 
+const MAX_UINT_64_STR = "18446744073709551615"
 // Max supply in lovelace
 const MAX_LOVELACE_SUPPLY_STR = ["45", "000", "000", "000", "000000"].join("");
+const POOL_MARGIN_DENOMINATOR_MAX = ["1", "000", "000", "000", "000", "000000"].join("")
```
janmazak, line 21: This could have STR in the name, and I guess we should use MAX consistently (I use _MAX in the end in C, following UINT64_MAX etc.).

janmazak (approved):

PR #32, opened 2021-02-11
Proper linting

ppershing:


.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+lib
+tests/lib
```
refi93, line 3:
```suggestion
test/lib
example-node/lib
```

.eslintrc
```diff
@@ -0,0 +1,47 @@
...
+  "plugins": ["flowtype", "import", "prettier", "mocha"],
+  "rules": {
+    "no-unused-vars": [
+      "warn",
+      {
+        "varsIgnorePattern": "^_",
+        "argsIgnorePattern": "^_"
+      }
+    ],
+    "flowtype/boolean-style": [2, "boolean"],
```
refi93, line 18: a bit of a nitpick but would be perhaps nice to replace the numerical constants with its verbal equivalents - warn/error for the sake of clarity

workspace.code-workspace
```diff
@@ -0,0 +1,11 @@
+{
+	"folders": [
+		{
+			"path": "."
```
refi93, line 4: don't we want to exclude lib/node modules so it doesn't slow search down/pollute it with meaningless results?

.eslintignore
refi93: the original patterns, especially "lib" does not work probably as intended as it matches any file/folder with that name regardless of where in the file tree it actually is: https://www.atlassian.com/git/tutorials/saving-changes/gitignore

refi93 (changes requested):
a few improvement suggestions and also `yarn lint` fails with two automatically fixable errors in `/test/src/integration/getExtendedPublicKey.js` - missing semicolon

The rest of the eslint/flow errors probably makes sense to tackle in a separate PR

.eslintrc
ppershing: fixed

workspace.code-workspace
ppershing: I would say - if you have any workspace preferences, go for it as a followup PR.

ppershing:
> a few improvement suggestions and also `yarn lint` fails with two automatically fixable errors in `/test/src/integration/getExtendedPublicKey.js` - two missing semicolons
> 
> The rest of the eslint/flow errors probably makes sense to tackle in a separate PR

For now I will be skipping tests (as long as they work). They will need a huge pass once we switch to TS anyway ...

refi93 (approved):

PR #33, opened 2021-02-12
First phase of migration to TS - move src/ folder and yolo fix immediate errors.

ppershing:
Note: This was just a brute-force of minimal type adjustments for the code to compile. The typing is terrible and we will work on that. In particular, we will have to cleanup `optional?: type` vs `value: type | undefined` vs `value: type | null` and make one coherent usage.

### DIFF
--- a/src/cardano.ts
+++ b/src/cardano.ts
@@ -1,5 +1,4 @@
-//@flow
-import utils, { Precondition, Assert } from "./utils";
+import utils, { Precondition, Assert, invariant } from "./utils";
 import type {
   TxOutput,
   Certificate,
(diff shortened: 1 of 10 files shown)


src/cardano.ts
```diff
@@ -1,5 +1,5 @@
 //@flow
```
refi93, line 1: I guess we can remove this

src/cardano.ts
```diff
@@ -124,7 +125,7 @@ function validateCertificates(certificates: Array<Certificate>) {
           !!poolParams,
           TxErrors.CERTIFICATE_POOL_MISSING_POOL_PARAMS
         );
-
+        invariant(!!poolParams)
         // serialization succeeds if and only if params are valid
         // $FlowFixMe
```
refi93, line 130: we can probably get rid of those FlowFixMe comments altogether if this is already a ts file

refi93 (approved):
LGTM, just a few redundancies we can already get rid of IMHO

PR #34, opened 2021-02-12
Setup continuous integration

refi93:
Motivation: setup continuous integration that would verify the project is building properly, run style checks and audit dependencies for known vulnerabilities.

Testing:
* tested locally by running `circleci local execute --job build_and_test` - fails during "Lint" stage as expected given the ongoing refactor


ppershing:
seems legit

PR #36, opened 2021-02-12
Split out Ledger interaction flows to separate files

ppershing:


src/Ada.ts
```diff
@@ -229,32 +220,6 @@ export const getErrorDescription = (statusCode: number) => {
...
-      ) {
-        // Do the retry
-        return await fn(...args);
-      }
-      throw e;
-    }
-  };
-}
-
 function wrapConvertError<T extends Function>(fn: T): T {
```
refi93, line 223: don't we want to extract this one as well?

src/interactions/ins.ts
```diff
@@ -0,0 +1,11 @@
+export const enum INS {
```
refi93, line 1: it kind of feels weird to have this alongside the actual "interactions" - perhaps I would put it in a "common" subfolder alongside `utils.ts` What do you think?

refi93:
two minor suggestions, otherwise LGTM

src/interactions/ins.ts
```diff
@@ -0,0 +1,11 @@
...
+  GET_VERSION = 0x00,
+  GET_SERIAL = 0x01,
+
+  GET_EXT_PUBLIC_KEY = 0x10,
+  DERIVE_ADDRESS = 0x11,
+
+  SIGN_TX = 0x21,
+
+  RUN_TESTS = 0xf0,
+}
```
refi93, line 11: newline at the end

src/interactions/ins.ts
ppershing: we can argue whether this should go into separate subfolder or not, but it definitely belongs alongside interactions as it is "I/O" definition. For now I will leave it as it is, there is plenty of other refactoring to do and in the process I might move it around anyway

src/Ada.ts
ppershing: Not necessarily. I did not want to extract the first function but I had no choice to avoid circular imports. Once we get rid of sendFn dependency in interactions, we will move it back here

PR #37, opened 2021-02-12
Fix regression introduced by types lying about undefined vs null.

ppershing:
Note that this is just a hotfix and the whole typing will be refactored in the future.

### DIFF
--- a/src/cardano.ts
+++ b/src/cardano.ts
@@ -205,7 +205,7 @@ export function validateTransaction(
     // we try to serialize the data, an error is thrown if ada amount or address params are invalid
     serializeOutputBasicParams(output, protocolMagic, networkId);
 
-    if ('spendingPath' in output) {
+    if ("spendingPath" in output && output.spendingPath != null) {
       Precondition.check(
         !isSigningPoolRegistrationAsOwner,
         TxErrors.OUTPUT_WITH_PATH
@@ -393,7 +393,7 @@ export function serializeOutputBasicParams(
   let outputType;
   let addressBuf;
 
-  if ('addressHex' in output) {
+  if ("addressHex" in output && output.addressHex) {
     outputType = TxOutputTypeCodes.SIGN_TX_OUTPUT_TYPE_ADDRESS_BYTES;
 
     Precondition.checkIsHexString(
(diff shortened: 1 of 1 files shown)


src/cardano.ts
```diff
@@ -606,7 +606,7 @@ export function serializePoolRelayParams(relayParams: RelayParams): Buffer {
   noBuf.writeUInt8(RELAY_NO);
 
   let portBuf: Buffer;
-  if ('portNumber' in params) {
+  if ('portNumber' in params && params.portNumber) {
```
refi93, line 609:
```suggestion
  if ('portNumber' in params && params.portNumber != null) {
```

src/cardano.ts
```diff
@@ -670,7 +670,7 @@ export function serializePoolRelayParams(relayParams: RelayParams): Buffer {
   }
 
   let dnsBuf: Buffer | undefined;
-  if ('dnsName' in params) {
+  if ("dnsName" in params && params.dnsName) {
```
refi93, line 673:
```suggestion
  if ("dnsName" in params && params.dnsName != null) {
```

src/cardano.ts
refi93: technically, portnumber can be 0

src/cardano.ts
refi93: teeechnically, dns name can be empty string

src/cardano.ts
```diff
@@ -205,7 +205,7 @@ export function validateTransaction(
     // we try to serialize the data, an error is thrown if ada amount or address params are invalid
     serializeOutputBasicParams(output, protocolMagic, networkId);
 
-    if ('spendingPath' in output) {
+    if ("spendingPath" in output && output.spendingPath != null) {
```
refi93, line 208: is there any point in checking whether given property is in the object if we are doing the `!= null` comparison? I know this is a hotfix, just asking if I'm perhaps missing something

src/cardano.ts
refi93: indeed, I noticed the original code has the same problem

refi93 (approved):

src/cardano.ts
ppershing: The check is covered, but Typescript does not like checks on maybe-undefined properties (especially the ones coming from union types)

src/cardano.ts
ppershing: Yep. This condition allows for empty dns name to pass. We can deal with the superexact semantic (can be empty) later

PR #38, opened 2021-02-13
Convert AddressTypeNibble/CertificateType to enums

ppershing:


test/src/test_utils.js
```diff
@@ -1,7 +1,8 @@
 import "babel-polyfill";
+
 import TransportNodeHid from "@ledgerhq/hw-transport-node-hid";
 
-import Ada, { utils, cardano } from "../../lib/Ada";
+import Ada, { cardano,utils } from "../../lib/Ada";
```
refi93, line 5:
```suggestion
import Ada, { cardano, utils } from "../../lib/Ada";
```
any idea why didn't the linter catch/fix it?

refi93: ok, comma-spacing rule is apparently missing in .eslintrc for .js files - I guess there is not much to gain in enabling it if we are migrating away from it anyway :)

for ts is apparently implicitly enabled by some of the plugins by default as I tried violating it in a ts file and the linter caught it

refi93 (approved):
LGTM

PR #40, opened 2021-02-13
Refactor: Extract sendTx sub-functions

ppershing:


src/interactions/signTx.ts
```diff
@@ -23,465 +24,496 @@ const PoolRegistrationCodes = {
...
 
-  const appHasStakePoolOwnerSupport = await isLedgerAppVersionAtLeast(
-    _send,
-    2,
-    1
-  );
-  if (isSigningPoolRegistrationAsOwner && !appHasStakePoolOwnerSupport) {
-    throw Error(Errors.INCORRECT_APP_VERSION);
-  }
+const P1_STAGE_INIT = 0x01;
```
refi93, line 28: maybe worth introducing an _InsFirstParam and _InsSecondParam auxiliary types which would be required in SendFn which would give it a more explicit interface - are you planning something similar or do you have some different approach in mind?

refi93 (approved):

src/interactions/signTx.ts
```diff
@@ -23,465 +24,496 @@ const PoolRegistrationCodes = {
...
-      }
+    return utils.uint8_to_buf(
+      isSigningPoolRegistrationAsOwner
+        ? PoolRegistrationCodes.SIGN_TX_POOL_REGISTRATION_YES
+        : PoolRegistrationCodes.SIGN_TX_POOL_REGISTRATION_NO
+    );
+  };
 
+  const _serializeIncludeInTxData = (
+    hasMultiassetSupport: boolean
```
refi93, line 77: for the sake of consistency, shouldn't this be:

```
flags: {
  appHasMultiassetSupport: boolean
}
```
?

But in general the whole function is a bit cringy, relying on outer-scope parameters and behavior governed by a single flag - kind of begs for splitting into two separate pure function which I don't insist on doing now

src/interactions/signTx.ts
ppershing: P1 will probably go as enum in the end. I am not sure I want to bind explicit enum to the function interface though. It is very unlikely that you will confuse the parameter with something else

src/interactions/signTx.ts
ppershing: Yeah. I think some other refactoring pass will clean this up. In any case, I wanted to introduce flags mainly to the extremely huge-argument-list outer functions to keep sanity. (It is more safe to pass booleans via {xyz} and then destructure to xyz than remembering correctly if you have xyz, abc, or efg and matching them correctly, especially because the names can be easily confused).

PR #42, opened 2021-02-14
More SignTx cleanups

ppershing:


src/interactions/signTx.ts
```diff
@@ -270,53 +297,53 @@ const signTx_addCertificate = async (
...
 
     // additional data for pool certificate
-    const APDU_INSTRUCTIONS = {
-      POOL_PARAMS: 0x30,
-      OWNERS: 0x31,
-      RELAYS: 0x32,
-      METADATA: 0x33,
-      CONFIRMATION: 0x34,
-    };
+    const enum P2 {
```
refi93, line 300: I didn't really know redeclaring const within a smaller scope is a thing :D Wondering if it wouldn't make sense to declare the P2 enum for unused P2 as a top-level const instead of "tediously" redeclaring it for each sub-instruction

ppershing: it woudln't. P2 makes sense only in the context of specific P1 command. I would rather re-declare than to mix random P2_UNUSED with P2_something

refi93: but at the same time I can see it getting awkward if we had some 2-level nested scopes (which I believe we should not really have here, even if Ledger itself had such level of nesting in their state machines, as we can break it down into independent functions)

refi93 (approved):
LGTM

src/interactions/signTx.ts
refi93: Fair point, I've written the comment above before seeing your response

PR #43, opened 2021-02-14
Simplyfy checking of fixed-length hex strings

ppershing:


src/utils.ts
```diff
@@ -64,12 +64,15 @@ export const Precondition = {
...
   },
-
   checkIsHexString: (data: any, msg?: string) => {
     Precondition.checkIsString(data, msg);
     Precondition.check(data.length % 2 === 0, msg);
     Precondition.check(/^[0-9a-fA-F]*$/.test(data), msg);
   },
+  checkIsHexStringOfLength: (data: any, expectedByteLength: number, msg?: string) => {
+    Precondition.checkIsHexString(data, msg)
+    Precondition.check(data.length == expectedByteLength * 2, msg)
```
refi93, line 74: any reason to use `==` instead of `===`?

refi93 (approved):
LGTM, one minor remark

src/utils.ts
ppershing: not really. We don't have eslint rule for this yet so I missed it

PR #44, opened 2021-02-15
Split validation&serialization for Pool relay

ppershing:


src/cardano.ts
```diff
@@ -565,127 +568,150 @@ export function serializePoolOwnerParams(params: PoolOwnerParams): Buffer {
...
-        ipPart,
-        TxErrors.CERTIFICATE_POOL_RELAY_INVALID_IPV4
-      );
-      ipBytes.writeUInt8(ipPart, i);
-    }
-    ipv4Buf = Buffer.concat([yesBuf, ipBytes]);
-  } else {
-    ipv4Buf = noBuf;
-  }
+// FIXME(ppershing): This is terrible and wrong implementation
```
refi93, line 611: do you see any particular issue with this implementation apart from not supporting abbreviated ipv6 addresses? It's a fair point though that the canonical representation of ipv6 is abbreviated but introducing a whole library/non-trivial helper function for this didn't seem to justify the benefits as manipulation with IP addresses is not the main purpose of this library and for the caller it should be very simple to expand the address before passing it to ledgerjs which is also documented: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/develop/docs/certificates.md#single-host-ip

refi93: ok, another obvious issue of this implementation even if we settle for full ipv6 addresses is that you can pass a string with randomly scattered ":" and it would fly as a valid ipv6 address

ppershing: Not to mention that even full ipv6 addresses usually skip leading zeros

src/cardano.ts
```diff
@@ -565,127 +568,150 @@ export function serializePoolOwnerParams(params: PoolOwnerParams): Buffer {
...
-  Precondition.checkIsUint8(type);
-
-  let typeBuf = Buffer.alloc(1);
-  typeBuf.writeUInt8(type);
-
-  switch (type) {
-    case 0:
-      return Buffer.concat([typeBuf, portBuf, ipv4Buf, ipv6Buf]);
+export function serializePoolRelay(relay: ParsedPoolRelay) {
+  function optional<T>(x: T | null, cb: (t: T) => Buffer): Buffer {
```
refi93, line 674: I would choose a less generic name to make it clear this "optional" is really specific only to this codebase if not function. At first I kind of thought this is referring to some language-level concept of optionality in typescript (as there are optionals e.g. in Java), so I'd better write it in a way so others without affirmative knowledge of TypeScript don't fall for this "fallacy"

Perhaps - serializeOptional ? And the "Optional" enum could be e.g. "OptionalCode" for that purpose.

refi93 (approved):
one question and one minor remark, otherwise LGTM

src/cardano.ts
```diff
@@ -501,230 +494,318 @@ export function serializePoolInitialParams(params: PoolParams): Buffer {
...
 
-  return Buffer.concat([
-    includeMetadataBuffer,
-    utils.hex_to_buf(hashHex),
-    Buffer.from(url),
-  ]);
+export function serializePoolMetadata(
+  metadata: ParsedPoolMetadata | null
+): Buffer {
+  if (metadata == null) {
```
refi93, line 798: I would consider generalizing the `serializeOptional` function in `serializePoolRelay()` to be applicable here (and eventually anywhere else where we deal with optional data

src/cardano.ts
```diff
@@ -501,230 +494,318 @@ export function serializePoolInitialParams(params: PoolParams): Buffer {
...
+  SingleHostAddr = 0,
+  SingleHostName = 1,
+  MultiHostName = 2,
+}
 
-  const noBuf = Buffer.alloc(1);
-  noBuf.writeUInt8(RELAY_NO);
+type ParsedPoolRelay = {
+  type: RelayType.SingleHostAddr,
+  port: number | null,
```
refi93, line 633: based on the recent discussion with IOHK currently pools with no port/ips get disregarded by the chain even though the transaction doesn't get rejected and for the pool operator app we will update the logic to enforce the port number and one or both of the ipv4 and ipv6 address.

The port can technically be null, but currently such pool is not valid and in the future it may have some default meaning which was not decided yet - so no point in allowing null port either.

So imho there is no harm in already enforcing a proper SingleHostAddr relay at js level even though the ledger app logic is more relaxed.

refi93 (approved):
a few minor remarks on the newly pushed changes

src/cardano.ts
ppershing: I think the constants are already different so it gets a bit hard to generalize. In either case, you can go for generalization if you feel like it. In my first pass I want to first get reasonable structure of the code, we can deal with polishing local details later...

PR #48, opened 2021-02-19
Cleanup certificates parsing & serialization

ppershing:


src/cardano.ts
```diff
@@ -60,86 +60,80 @@ export const GetKeyErrors = {
...
+function parseCertificate(cert: Certificate): ParsedCertificate {
+  switch (cert.type) {
+    case CertificateType.STAKE_REGISTRATION:
+    case CertificateType.STAKE_DEREGISTRATION: {
+      Precondition.checkIsValidPath(
+        cert.path,
+        TxErrors.CERTIFICATE_MISSING_PATH
+      );
+      Precondition.check(
+        !cert.poolKeyHashHex,
```
refi93, line 87: this should probably have been `cert.poolKeyHashHex != null` as it would not trigger for the key being a falsy value like an empty string or zero

src/interactions/signTx.ts
```diff
@@ -250,105 +250,114 @@ const signTx_addOutput = async (
...
+        expectedResponseLength: 0,
+      });
+      break
+    }
     case CertificateType.STAKE_DELEGATION: {
-      if (certificate.path != null) dataFields.push(utils.path_to_buf(certificate.path));
+      const data = Buffer.concat([
+        utils.uint8_to_buf(certificate.type),
+        utils.path_to_buf(certificate.path),
+        utils.hex_to_buf(certificate.poolKeyHashHex)
```
refi93, line 279: nitpick - sometimes the comma is dangling, sometimes not, I'd add a rule to enforce it for consistency: https://eslint.org/docs/rules/comma-dangle

refi93 (approved):
integration tests passed, LGTM, only two minor remarks

src/cardano.ts
ppershing: actually, it should have been == null

src/interactions/signTx.ts
ppershing: For some reason, autoformat is still not perfect. I will have to investigate but formatting is not trying to re-wrap multi-line expressions/call sites and I am therefore not sure whether prettier kicks in correctly

PR #49, opened 2021-02-20
Split validation & serialization

ppershing:


src/interactions/signTx.ts
```diff
@@ -9,10 +9,18 @@ import type {
   Withdrawal,
 } from "../Ada";
 import {
-  CertificateType,
   Errors,
 } from "../Ada"
-import cardano, { ParsedCertificate, SignTxIncluded } from "../cardano";
+import cardano, { SignTxIncluded } from "../cardano";
+import type {
```
refi93, line 15: can't those three import statements be merged into a single one as they are apparently importing from the same location?

src/parsing.ts
```diff
@@ -0,0 +1,513 @@
...
+    SingleHostIPRelay,
+    SingleHostNameRelay,
+    TxOutputTypeAddress,
+    TxOutputTypeAddressParams,
+    Withdrawal,
+} from "./Ada";
+// TODO: remove this dependency
+import { serializeOutputBasicParams } from "./cardano";
+import { TxErrors } from "./txErrors";
+import utils, { MAX_LOVELACE_SUPPLY_STR, Precondition } from "./utils";
```
refi93, line 20: those two imports can also be merged, maybe worth enabling no-duplicate-imports rule https://eslint.org/docs/rules/no-duplicate-imports

src/cardano.ts
```diff
@@ -806,38 +304,24 @@ export function serializePoolMetadata(
   }
 }
 
-export function serializeGetExtendedPublicKeyParams(path: BIP32Path): Buffer {
-  Precondition.check(!!path, GetKeyErrors.INVALID_PATH);
-
-  return utils.path_to_buf(path);
+export function serializeGetExtendedPublicKeyParams(path: ValidBIP32Path): Buffer {
+  return Buffer.concat([
+    utils.path_to_buf(path),
```
refi93, line 309: why concatenating a single buffer? Isn't it equivalent to returning `utils.path_to_buf(path)` directly?

src/parsing.ts
```diff
@@ -0,0 +1,513 @@
...
+        Precondition.checkIsPositiveUint64Str(
+            validityIntervalStartStr,
+            TxErrors.VALIDITY_INTERVAL_START_INVALID
+        );
+    }
+}
+
+
+
+export type VarlenAsciiString = string & { __type: 'ascii' }
```
refi93, line 238: nitpick - I would prefer to keep types visually separated from the functions - e.g. types at the top and functions at the bottom, especially types such as FixlenHexString or ValidBIP32Path look generic enough to not be randomly introduced in the middle of the file

src/parsing.ts
```diff
@@ -0,0 +1,513 @@
...
+        default:
+            throw new Error(TxErrors.CERTIFICATE_INVALID);
+    }
+}
+
+export function parseCertificates(certificates: Array<Certificate>): Array<ParsedCertificate> {
+    Precondition.checkIsArray(certificates, TxErrors.CERTIFICATES_NOT_ARRAY);
+
+    const parsed = certificates.map(cert => parseCertificate(cert))
+
```
refi93, line 114: redundant empty line

src/cardano.ts
refi93: if the function is that simple and it's used one, I would consider putting the `utils.path_to_buf(path)` call right where the function is invoked

src/parsing.ts
```diff
@@ -0,0 +1,513 @@
...
+    PoolMetadataParams,
+    PoolOwnerParams,
+    PoolParams,
+    RelayParams,
+    SingleHostIPRelay,
+    SingleHostNameRelay,
+    TxOutputTypeAddress,
+    TxOutputTypeAddressParams,
+    Withdrawal,
+} from "./Ada";
```
refi93, line 16: this makes for a cyclic import as Ada.ts imports the AddressTypeNibble and CertificateType from here - wouldn't it make sense to keep those types within Ada.ts as they are relevant for the library users as well?

src/parsing.ts
```diff
@@ -0,0 +1,513 @@
...
+}
+
+export enum CertificateType {
+    STAKE_REGISTRATION = 0,
+    STAKE_DEREGISTRATION = 1,
+    STAKE_DELEGATION = 2,
+    STAKE_POOL_REGISTRATION = 3,
+}
+
+const KEY_HASH_LENGTH = 28;
```
refi93, line 38: are we ok with keeping the duplicate of this constant in cardano.ts? is this just some temporary stage towards further reorganization of types/constants?

src/cardano.ts
```diff
@@ -47,206 +21,6 @@ export const SignTxIncluded = Object.freeze({
 const KEY_HASH_LENGTH = 28;
```
refi93, line 21: not sure if left out by mistake or on purpose but duplicates of KEY_HASH_LENGTH and TX_HASH_LENGTH are declared in parsing.ts

refi93 (changes requested):
see comments - not really significant remarks, I'm mostly concerned about apparently duplicate constants and potentially better places for some types

src/parsing.ts
refi93: ok, not really, probably it's better like it is now, in the end, the types give additional insight when reading code that uses them

ppershing: It is likely that the basic types + parsing functions eventually move to helpers anyway

src/parsing.ts
ppershing: Don't get me even start on the import cycles. I know. But that is for another type. We need quite a few restructurings, especially splitting external & internal types, etc.

src/interactions/signTx.ts
ppershing: Type imports cannot be mixed with normal imports. But yes, the second two can be merged

ppershing:
Fixed most of the things

PR #50, opened 2021-02-21
Continue refactor + parse asset group

ppershing:


src/parsing.ts
```diff
@@ -510,3 +549,174 @@ export function parsePoolMetadataParams(params: PoolMetadataParams | null): Pars
...
+            stakingChoice: StakingChoiceNone // included in spending path
+        }
+    )
+
+export type ParsedAddressParams = ByronAddressParams | ShelleyAddressParams
+
+export function parseAddressParams(
+    params: AddressParams
+): ParsedAddressParams {
+    if (params.addressTypeNibble === AddressTypeNibble.BYRON) {
```
refi93, line 617: any specific reason to leave out byron addresses out of the switch statement below?

src/parsing.ts
```diff
@@ -120,6 +123,41 @@ export function parseCertificates(certificates: Array<Certificate>): Array<Parse
...
+
+export type ParsedAssetGroup = {
+    policyIdHex: FixlenHexString<typeof TOKEN_POLICY_LENGTH>,
+    tokens: Array<ParsedToken>,
+};
+
+function parseToken(token: Token): ParsedToken {
+    const assetNameHex = parseHexString(token.assetNameHex, TxErrors.OUTPUT_INVALID_ASSET_NAME);
+    Precondition.check(
+        token.assetNameHex.length <= TOKEN_NAME_LENGTH * 2,
```
refi93, line 140: this is preexisting code, but there's an apparent discrepancy between the constant name and property name - I would rename the constant to `MAX_ASSET_NAME_LENGTH`

src/cardano.ts
```diff
@@ -126,7 +74,7 @@ export function serializeOutputBasicParams(
   let addressBuf;
 
   if ("addressHex" in output && output.addressHex) {
```
refi93, line 76: maybe worth doing in another pass of the refactor but begs for being rewritten to output.addressHex != null and similarly for other similar instances - found three in the current code by looking up `" in` pattern

refi93 (approved):
LGTM, two minor remarks

PR #51, opened 2021-02-23
Refactor: split TxOutput validation & serialization

ppershing:


src/parsing.ts
```diff
@@ -720,4 +722,67 @@ export function parseAddressParams(
...
+}
+
+export function parseTxOutput(
+    output: TxOutput,
+    network: Network,
+): ParsedOutput {
+    const amountStr = parseUint64_str(output.amountStr, MAX_LOVELACE_SUPPLY_STR, TxErrors.OUTPUT_INVALID_AMOUNT)
+    const assets = (output.tokenBundle ?? []).map((ag) => parseAssetGroup(ag))
+
+    const hasAddressHex = "addressHex" in output && output.addressHex != null
```
refi93, line 749: isn't the `"addressHex" in output` check redundant? same for `spendingPath` below

ppershing: No, this is typescript limitation with TxOutput being typed as union with one variant missing the key

src/parsing.ts
```diff
@@ -720,4 +722,67 @@ export function parseAddressParams(
...
+    assets: AssetGroup[]
+    destination: OutputDestination
+}
+
+export function parseTxOutput(
+    output: TxOutput,
+    network: Network,
+): ParsedOutput {
+    const amountStr = parseUint64_str(output.amountStr, MAX_LOVELACE_SUPPLY_STR, TxErrors.OUTPUT_INVALID_AMOUNT)
+    const assets = (output.tokenBundle ?? []).map((ag) => parseAssetGroup(ag))
```
refi93, line 747: I'd keep the "tokenBundle" name to stay consistent with the external interface and established terminology (see https://developers.cardano.org/en/development-environments/native-tokens/native-tokens-glossary/). Also, I find it a bit awkward to have something named "assets" only to assign to it "asset groups"

refi93 (approved):
one minor naming remark, otherwise LGTM

PR #52, opened 2021-02-24
Refactor withdrawal parsing

ppershing:


src/interactions/signTx.ts
```diff
@@ -615,8 +613,8 @@ export async function signTransaction(
   }
 
   if (withdrawals.length > 0) {
```
refi93, line 615: isn't this condition redundant?

refi93 (approved):
LGTM, one minor remark

PR #57, opened 2021-03-06
Refactor: split & move types

ppershing:


src/types/public.ts
```diff
@@ -0,0 +1,155 @@
+import type { AddressTypeNibble } from './internal'
+
+export { AddressTypeNibble } from './internal'
```
refi93, line 3: is it worth keeping `AddressTypeNibble` among internal types if it is exported from here anyway? I'd find it more intuitive to use external types within internal

src/types/public.ts
```diff
@@ -0,0 +1,155 @@
...
+
```
refi93, line 4: Also, I'd export the CertificateType enum here as it may be handy for integrating parties, Yoroi uses it currently: If we export the CertificateType enum si

refi93: and same with `RelayType` for the sake of completeness

refi93 (approved):
LGTM, I'd just make "public" a few more auxiliary types which may come in handy for integrating parties, see comments

src/types/public.ts
ppershing: :thinking: good point, makes sense

PR #58, opened 2021-03-06
Pass explicitly version to interactions

ppershing:


src/Ada.ts
```diff
@@ -313,7 +318,8 @@ export default class Ada {
       stakingBlockchainPointer
     })
 
-    return showAddress(this._send, addressParams);
+    const version = await getVersion(this._send)
```
refi93, line 321: to make it clear what kind of version are we referring to (one may wonder if there can be multiple "versions" of the same call), I'd name the variable appVersion

src/interactions/getExtendedPublicKeys.ts
```diff
@@ -12,11 +12,12 @@ import { ensureLedgerAppVersionAtLeast } from "./getVersion";
 
 export async function getExtendedPublicKeys(
   _send: SendFn,
+  version: Version,
```
refi93, line 15: for the sake of consistency with other interactions, this should be prefixed with an underscore

src/interactions/getSerial.ts
```diff
@@ -1,12 +1,13 @@
...
+import type { Version } from "../types/internal"
 import type { GetSerialResponse } from "../types/public"
 import utils, { assert } from "../utils";
 import { INS } from "./common/ins";
 import { wrapRetryStillInCall } from "./common/retry";
 import { ensureLedgerAppVersionAtLeast } from "./getVersion";
 
-export async function getSerial(_send: SendFn): Promise<GetSerialResponse> {
-  await ensureLedgerAppVersionAtLeast(_send, 1, 2);
+export async function getSerial(_send: SendFn, version: Version): Promise<GetSerialResponse> {
```
refi93, line 9: other interactions have the version parameter underscored

ppershing: it is underscored if it is unused

src/Ada.ts
ppershing: noted but for now I will leave it as it is

src/interactions/getSerial.ts
refi93: I see

PR #61, opened 2021-03-07
Switch api

ppershing:


src/Ada.ts
```diff
@@ -205,15 +203,14 @@ export default class Ada {
   /**
    * @description Get several public keys; one for each of the specified BIP 32 paths.
    *
-   * @param {Array<BIP32Path>} paths The paths. A path must begin with `44'/1815'/account'` or `1852'/1815'/account'`, and may be up to 10 indexes long.
-   * @return {Promise<Array<GetExtendedPublicKeyResponse>>} The extended public keys (i.e. with chaincode) for the given paths.
+   * @param paths The paths. A path must begin with `44'/1815'/account'` or `1852'/1815'/account'`, and may be up to 10 indexes long.
+   * @returns The extended public keys (i.e. with chaincode) for the given paths.
```
refi93, line 207:
```suggestion
   * @return The extended public keys (i.e. with chaincode) for the given paths.
```

src/Ada.ts
```diff
@@ -280,81 +277,37 @@ export default class Ada {
...
-    protocolMagic: number,
-    inputs: Array<InputTypeUTxO>,
-    outputs: Array<TxOutput>,
-    feeStr: string,
-    ttlStr: string | null,
-    certificates: Array<Certificate>,
-    withdrawals: Array<Withdrawal>,
-    metadataHashHex?: string | null,
-    validityIntervalStartStr?: string | null
+    tx: Transaction
```
refi93, line 300: I would name this `SignTransactionRequest` - a bit more eloquent but it matches up with the response type naming and I can see the "Transaction" name becoming misleading as soon as we would for whatever reason need to pass some auxiliary flags to the signTransaction call that wouldn't really be the Transaction itself

src/Ada.ts
```diff
@@ -280,81 +277,37 @@ export default class Ada {
    *
    */
   async deriveAddress(
-    addressTypeNibble: AddressTypeNibble,
-    networkIdOrProtocolMagic: number,
-    spendingPath: BIP32Path,
-    stakingPath: BIP32Path | null = null,
-    stakingKeyHashHex: string | null = null,
-    stakingBlockchainPointer: StakingBlockchainPointer | null = null
+    addressParams: AddressParams
```
refi93, line 280: for the sake of consistency, I'd name the types for the input objects with the convention CallNameRequest and in this case I'd create an `DeriveAddressRequest` type alias for this parameter as I see rather as a coincidence that it's enough to pass directly the AddressParams object

src/Ada.ts
```diff
@@ -205,15 +203,14 @@ export default class Ada {
...
+   * @returns The extended public keys (i.e. with chaincode) for the given paths.
    *
    * @example
    * const [{ publicKey, chainCode }] = await ada.getExtendedPublicKeys([[ HARDENED + 44, HARDENED + 1815, HARDENED + 1 ]]);
    * console.log(publicKey);
    *
    */
-
   async getExtendedPublicKeys(
     paths: Array<BIP32Path>
```
refi93, line 215: I'd go all in with object based inputs and wrap those paths in a `GetExtendedPublicKeysRequest` to keep a consistent integration experience across the calls offered

src/Ada.ts
```diff
@@ -125,7 +125,7 @@ export type SendFn = (params: SendParams) => Promise<Buffer>;
 
 export type Transport = any
 
-export default class Ada {
+export class Ada {
```
refi93, line 128: I noticed the methods listing below has the `getExtendedPublicKey` call - I think it should actually be `getExtendedPublicKeys` so the `decorateAppAPIMethods` actually applies the concurrent execution protection

src/Ada.ts
ppershing: Well, go figure, half of the methods are annotated with `@returns` and half with `@return` :man_shrugging:

refi93: I see, from the surrounding changes I got the impression return is the correct way but both are apparently used interchangeably: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

src/parsing.ts
```diff
@@ -265,8 +252,8 @@ export function parsePoolParams(params: PoolParams): ParsedPoolParams {
     return {
         keyHashHex,
         vrfHashHex,
-        pledgeStr,
-        costStr,
+        pledge: pledge,
```
refi93, line 255:
```suggestion
        pledge,
        cost,
```

refi93, line 255:
```suggestion
        pledge,
        cost,
```

src/parsing.ts
```diff
@@ -265,8 +252,8 @@ export function parsePoolParams(params: PoolParams): ParsedPoolParams {
...
+        cost: cost,
```
refi93, line 256:
```suggestion
        pledge,
        cost,
```

src/types/public.ts
```diff
@@ -172,3 +235,17 @@ export type SignTransactionResponse = {
...
+export type Transaction = {
+    network: Network,
+    inputs: Array<TxInput>,
+    outputs: Array<TxOutput>,
+    fee: bigint_like,
+    ttl: bigint_like | null,
+    certificates: Array<Certificate>,
+    withdrawals: Array<Withdrawal>,
+    metadataHashHex?: string | null,
+    validityIntervalStart?: string | null
```
refi93, line 250: this should be `bigint_like | null` I guess

src/parseUtils.ts
```diff
@@ -82,9 +88,20 @@ export function parseHexStringOfLength<L extends number>(str: unknown, length: L
...
 }
 
-export function parseUint64_str(str: unknown, constraints: { min?: string, max?: string }, err: string): Uint64_str {
-    validate(isUint64str(str) && isUintStr(str, constraints), err)
-    return str
+export function parseUint64_str(val: unknown, constraints: { min?: string, max?: string }, err: string): Uint64_str {
+    switch (typeof val) {
+        case 'string':
+            validate(isUint64str(val) && isUintStr(val, constraints), err)
+            return val
```
refi93, line 95: don't you need to cast this to Uint64_str as well?

ppershing: no, this is already guarded by isUint64str

PR #62, opened 2021-03-10
Drop backwards compatibility + introduce Compatibility api

ppershing:


refi93:
`getExtendedPublicKeys` function can also be cleaned up the bit from the backwards-compatibility adjustment for Ledger app 2.0.4, see https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/f9f486bc3658666acea0cf5272eedf9e9808b26f/src/interactions/getExtendedPublicKeys.ts#L45

refi93:
also, I'd align the `ensureLedgerAppVersionAtLeast` calls to version 2.2 or replace them with a check for compatibility - if we return that the given library is not compatible with the given device I find it a bit counterintuitive for any call but getVersion to get through

src/interactions/getVersion.ts
```diff
@@ -36,14 +35,35 @@ export function* getVersion(): Interaction<Version> {
   return { major, minor, patch, flags };
 }
 
+export function getCompatibility(version: Version): DeviceCompatibility {
+  const v2_2 = isLedgerAppVersionAtLeast(version, 2, 2) && isLedgerAppVersionAtMost(version, 2, 2)
+
+  return {
+    isCompatible: v2_2,
```
refi93, line 42: I'm a bit confused about the granularity of what being "compatible" here means - is it meant like some hard threshold for any call or should it be taken as a "gentle" warning that for any older version of the fw you would proceed at your own risk? I'd personally got with the former and enforce it for any call but getVersion so users don't shoot themselves in the foot

refi93 (approved):

PR #64, opened 2021-03-10
Cleanup error structure

ppershing:


src/parseUtils.ts
```diff
@@ -104,28 +104,28 @@ export function parseUint64_str(val: unknown, constraints: { min?: string, max?:
     }
 }
 
-export function parseUint32_t(value: unknown, err: string): Uint32_t {
+export function parseUint32_t(value: unknown, err: InvalidDataReason): Uint32_t {
```
refi93, line 107: the function signatures imho deserve to have the "err" parameter renamed to errMsg, as the "reason" isn't an error in its own right

refi93 (commented):
LGTM, one minor remark

PR #72, opened 2021-03-12
Setup docs building in CI

refi93:
Motivation: Don't commit autogenerated docs as it can generate unnecessary conflicts and commit the docs to a separated gh-pages branch instead from CI.

Moreover, allow for multiple versions of the docs (by package version number) to reside one besides another so it's easy to navigate between them. Unfortunately, typedoc doesn't seem to support versioning: https://github.com/TypeStrong/typedoc/issues/737

Explanation:
* created a CI job that builds the docs into a separate folder for the current version of the package
* added gh-pages development dependency in order to "deploy" the docs built into a dedicated gh-pages branch which is then hosted as a github page
* the docs reside in a folder named by the version number of the package. I added into CI a command to create an index.html file such that it redirects the visitor of the github page root to the latest version of the package. Users then can navigate to earlier versions by flipping the version number in the URL (I tried also generating a listing of files but my efforts led nowhere and the current approach seemed like an easy win)


package.json
```diff
@@ -71,7 +72,7 @@
     "clean": "bash ./script/clean.sh",
     "build": "bash ./script/build.sh",
     "watch": "bash ./script/watch.sh",
-    "gen-docs": "yarn typedoc",
+    "gen-docs": "yarn typedoc --out docs_generated",
```
ppershing, line 75: there is typedoc config where you can specify out

ppershing (approved):

PR #81, opened 2021-03-15
Bump package version

ppershing:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cardano-foundation/ledgerjs-hw-app-cardano",
-  "version": "2.2.0",
+  "version": "3.0.0",
```
refi93, line 3: I'd also update circleci to follow master branch where we merge after releasing so the docs don't get out of sync as we merge partial unreleased updates into develop

https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/1e6eff6e0fde8ab9b08577f693924c35613bf18b/.circleci/config.yml#L100

Later on, we may devise some smart approach to rebuild docs only when version number is bumped, but following master should be good enough for now

refi93: or let's do that once we are really done with the refactor and 3.0.0 release is ready

refi93: probably it's better to switch circleci to trigger for github tags

refi93 (approved):

PR #82, opened 2021-03-15
Unify enum casing

ppershing:


refi93:
found also this inconsistently cased enum, though internal: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/8f31a3815d1308300eca1d3a17c4595d4cff8a86/src/interactions/serialization/poolRegistrationCertificate.ts#L53

ppershing:
Fixed Optional enum

refi93 (approved):

PR #85, opened 2021-03-16
Slightly improve docs

ppershing:


src/types/public.ts
```diff
@@ -30,9 +50,25 @@ export enum AddressType {
...
+     */
     STAKE_REGISTRATION = 0,
+    /**
+     * Staking key deregistration certificate
+     * @see [[StakeDeregistrationParams]]
+     */
     STAKE_DEREGISTRATION = 1,
+    /**
+     * Stake delegation certificate
+     * @see [[StakeDeregistrationParams]]
```
refi93, line 65: `StakeDelegationParams` I guess

src/types/public.ts
```diff
@@ -235,9 +294,17 @@ export type TxOutput = {
...
+     * Destination is an address not owned by the device.
+     * Device will show this address in the review screen.
+     * @see [[ThirdPartyAddressParams]]
+     */
     THIRD_PARTY = 'third_party',
-    /** For address owned by the device. Device will not show the address (unles it seems fishy) */
+    /**
+     * Destination is an address owned by the device.
+     * Device will not show the address (unles it seems fishy)
+     * @see [[DeviceOwnedAddress]]
```
refi93, line 306: `ThirdPartyAddressParams` just embeds the addressHex, yet is suffixed with "Params", so I find it odd that `DeviceOwnedAddress` isn't called `DeviceOwnedAddressParams` as well

src/types/public.ts
```diff
@@ -531,7 +630,10 @@ export type DeviceCompatibility = {
  * @category Basic types
  */
 export type Serial = {
-    /** Serial number of the device. TODO: is this ascii? hex? */
+    /**
+     * Serial number of the device.
+     * Serial is 7 bytes represented here as 14-character hex string
```
refi93, line 635: Is that serial persistent after the device is wiped? I think not and it's probably worth clearing up in the docs, maybe you know for sure as you were implementing that feature, if I'm not mistaken @ppershing

src/types/public.ts
```diff
@@ -178,9 +226,13 @@ export type AddressParamsReward = {
  * @see [[Transaction]]
  */
 export type TxInput = {
-    /** UTxO hash */
+    /**
+     * UTxO hash
```
refi93, line 230: this sounds as if it was the hash of the utxo itself which is not correct, obviously, I'd reword it as `the hash of the transaction the UTxO is referencing`

src/types/public.ts
ppershing: @janmazak do you know?

src/types/public.ts
ppershing: We use `DeviceOwnedAddress` also in `getAddress`/`showAddress` calls and it sounds weird in that context with `Params` suffix

ppershing: Also, we basically have "Params" suffix for tagged union `params` type. E.g. We have Relay = tagged union, RelayXyzParams = union variants

src/parsing/poolRegistration.ts
```diff
@@ -19,7 +19,7 @@ import { hex_to_buf } from "../utils/serialize";
 import { MAX_LOVELACE_SUPPLY_STR, POOL_REGISTRATION_OWNERS_MAX, POOL_REGISTRATION_RELAYS_MAX } from "./constants";
 
 function parseMargin(params: PoolRegistrationParams['margin']): ParsedMargin {
-    const POOL_MARGIN_DENOMINATOR_MAX_STR = "1 000 000 000 000.000000".replace(/[ .]/, "")
+    const POOL_MARGIN_DENOMINATOR_MAX_STR = "1 000 000 000 000 000 000".replace(/[ .]/, "")
```
refi93, line 22:
```suggestion
    const POOL_MARGIN_DENOMINATOR_MAX_STR = "1 000 000 000 000 000 000".replace(/[ ]/, "")
```

src/types/public.ts
janmazak: I think it is not persistent. It was good enough for the purpose of distinguishing two connected devices, but it is not a truly unique serial number. 
https://www.reddit.com/r/ledgerwallet/comments/9dghpn/get_a_unique_identifier_of_a_ledger_nano_s/

refi93: I just tried it directly and the serial persisted even after the app being reinstalled and setting a different seed, so it's apparently a persistent value, hence I'd drop the comment about the serial non-persistence after restarting the device. I asked in ledger support channel to get further confirmation/clarification

refi93 (approved):
please remove comment on non-persistency of serial number after recent findings, otherwise LGTM

PR #86, opened 2021-03-16
Fix package distribution & change output dir to `dist`

ppershing:


tsconfig.base.json
```diff
@@ -3,6 +3,7 @@
     "allowJs": true,
     "allowSyntheticDefaultImports": true,
     "alwaysStrict": true,
+    "declaration": true,
```
refi93, line 6: question - does this flag enable exporting the type definitions? Does this mean we weren't exporting the types until now? Asking to understand if I need to make another pre-release so Adalite can make use of ledgerjs public types

refi93 (approved):

tsconfig.base.json
```diff
@@ -3,6 +3,7 @@
     "allowJs": true,
```
refi93, line 3: I guess we are fully migrated to ts by now, so if this flag was to allow some interoperability with legacy js code, we should be able to drop it

tsconfig.base.json
ppershing: Yep, exactly, we weren't generating .d.ts files

tsconfig.base.json
ppershing: done

PR #87, opened 2021-03-17
Introduce transaction metadata type

ppershing:


src/types/public.ts
```diff
@@ -690,6 +690,39 @@ export type SignedTransactionData = {
...
+
+/**
+ * Specifies transaction metadata.
+ * Metadata can be supplied to Ledger in multiple forms.
+ * @category Basic types
+ * @see [[TransactionMetadataType]]
+ */
+export type TransactionMetadata =
+    {
+        type: TransactionMetadataType.ArbitraryHash
```
refi93, line 713: with the introduction of the "voting registration metadata" we'll need to return in the response to signTransaction call also the voting registration signature.

I'm wondering if it would make sense to already return in the sign transaction response a "metadata" key which would contain an object with the hash for generic metadata and for the voting registration it would contain also an additional key with the voting registration signature

ppershing: That could be added in the future, it should be compatible to add new fields to response

refi93 (approved):

PR #92, opened 2021-03-24
Fix docs building, remove unused dependencies

refi93:
Motivation: CI failed with docs being built when a PR has been merged into develop. Docs building was triggered only in develop branch so it was not so easy to notice before. Also, the node-example was not being validated automatically, which the docs building command picked up for whatever reason. The proposed resolution is to verify independently the validity both of the node-example and docs which should prevent future issues with the development pipeline getting blocked after mergin a PR that originally passed CI.

Changes:
* fix compilation error in the example js file
* add an attempt to build and validate the example file into the CI pipeline
* add an attempt to build the docs into the CI pipeline to verify docs are building properly
* I noticed we no longer depend on flow/babel so I removed those dependencies
* removed unused scripts


script/clean.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-rm -rf lib/ flow-typed/
+rm -rf dist/
```
ppershing, line 5: I guess you can just move it to package.json completely, now that these scripts are one-liners

ppershing (approved):

script/clean.sh
refi93: good point

PR #96, opened 2021-03-31
Rename metadata to auxiliary data

refi93:
Motivation: reflect latest changes in Cardano's CDDL: https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl

Notes:
* renamed metadataHashHex to just hashHex as the purpose is fairly obvious from the encapsulating object type and to make it consistent with the way we pass the metadata hash in the pool registration params object
* I decided to keep the internal "STAGE_METADATA" enum intact as that's still how it's named in  the currently launched ledger app, I'd change it together with Catalyst support

Testing: ran `yarn test-integration` and it passed

### DIFF
--- a/src/errors/invalidDataReason.ts
+++ b/src/errors/invalidDataReason.ts
@@ -96,8 +96,8 @@ export enum InvalidDataReason {
   WITHDRAWAL_INVALID_AMOUNT = "invalid withdrawal amount",
   WITHDRAWAL_INVALID_PATH = "invalid withdrawal path",
 
-  METADATA_UNKNOWN_TYPE = "unknown metadata type",
-  METADATA_INVALID_HASH = "invalid metadata hash",
+  AUXILIARY_DATA_UNKNOWN_TYPE = "unknown auxiliary data type",
+  AUXILIARY_DATA_INVALID_HASH = "invalid auxiliary data hash",
 
   VALIDITY_INTERVAL_START_INVALID = "invalid validity interval start",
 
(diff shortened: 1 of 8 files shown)


src/errors/invalidDataReason.ts
```diff
@@ -96,8 +96,8 @@ export enum InvalidDataReason {
   WITHDRAWAL_INVALID_AMOUNT = "invalid withdrawal amount",
   WITHDRAWAL_INVALID_PATH = "invalid withdrawal path",
 
-  METADATA_UNKNOWN_TYPE = "unknown metadata type",
-  METADATA_INVALID_HASH = "invalid metadata hash",
+  AUXILIARY_DATA_UNKNOWN_TYPE = "unknown metadata type",
```
janmazak, line 99: Don't we want to change the msg too?

src/interactions/signTx.ts
```diff
@@ -1,7 +1,7 @@
-import type { ParsedCertificate, ParsedInput, ParsedMetadata, ParsedOutput, ParsedSigningRequest, ParsedTransaction, ParsedWithdrawal, Uint64_str, ValidBIP32Path, Version } from "../types/internal";
+import type { ParsedAuxiliaryData, ParsedCertificate, ParsedInput, ParsedOutput, ParsedSigningRequest, ParsedTransaction, ParsedWithdrawal, Uint64_str, ValidBIP32Path, Version } from "../types/internal";
 import { CertificateType, PoolOwnerType, TX_HASH_LENGTH } from "../types/internal";
 import type { SignedTransactionData } from '../types/public';
-import { TransactionMetadataType, TransactionSigningMode } from '../types/public';
+import { AuxiliaryDataType, TransactionSigningMode } from '../types/public';
```
janmazak, line 4: I would prefer to keep Transaction or Tx in the name, otherwise it is very general.

janmazak (approved):

Issue #94, opened 2021-03-29
Fix prettier - enforce consistent indentation 


refi93:
Currently we have files with 4 spaces per tab and 2 spaces per tab, this should be unified to 2 spaces per tab

janmazak:
Done in #98, 982f9d6c72b31a8f11df4f935cbc60d51e79d6a7, v3.2.0.

PR #98, opened 2021-04-06
Enforce consistent indentation of two spaces and no semicolons

### LINKED ISSUE #93: Fix prettier - enforce no semicolons

refi93:
Motivation: Currently we have files with 4 spaces per tab and 2 spaces per tab indentation which is messy. Turns out we weren't enforcing consistent indentation rule at all which is apparent from an inspection of the enabled rules I did on an arbitrary file

```yarn eslint --print-config src/parsing/poolRegistration.ts```

so the indentation was being kept consistent only within the individual files, likely by the editor itself.

source: https://stackoverflow.com/questions/63775411/is-there-a-way-to-list-all-active-eslint-prettier-rules-in-an-angular-project

Solution:
* explicitly enforce consistent indentation of 2 spaces in .eslintrc, of course this triggers a bunch of errors (like 2000) and would generate a sizeable diff resulting in many conflicts, so I'm leaving this there until we have no active workstream which could otherwise get broken by this, and then we should run `yarn lint --fix`, commit it and merge this

Note:
* not sure why dangling comma enforcement doesn't seem to apply to type exports, but the rest is enforced which is still sharply better than the current state

Testing: `yarn lint` triggers now errors related to inconsitent indentation, see CI checks status

closes #93 and #94

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
   "plugins": ["@typescript-eslint", "import", "simple-import-sort", "prettier"],
   "rules": {
     "indent": 2,
+    "semi": ["error", "never"],
     "simple-import-sort/imports": "warn",
     "no-unused-vars": "off",
     "@typescript-eslint/no-unused-vars": [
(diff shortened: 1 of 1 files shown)


janmazak:
Will be merged via operatorapp-ts.

PR #99, opened 2021-04-07
Catalyst voting

refi93:
WIP catalyst voting support


gabrielKerekes (approved):

src/interactions/getVersion.ts
```diff
@@ -38,12 +38,13 @@ export function* getVersion(): Interaction<Version> {
 export function getCompatibility(version: Version): DeviceCompatibility {
   // We restrict forward compatibility only to backward-compatible semver changes
   const v2_2 = isLedgerAppVersionAtLeast(version, 2, 2) && isLedgerAppVersionAtMost(version, 2, Infinity)
+  const v2_3 = isLedgerAppVersionAtLeast(version, 2, 3) && isLedgerAppVersionAtMost(version, 2, Infinity)
 
   return {
     isCompatible: v2_2,
-    recommendedVersion: v2_2 ? null : '2.2.0',
+    recommendedVersion: v2_2 ? null : '2.3.0',
```
janmazak, line 45: I don't understand the precise meaning of these two. Well, there is an explanatory comment at DeviceCompatibility definition, but it uses undefined term "SDK" and formulations like "true if SDK somewhat supports device". But that is insufficient for keeping this function up to date.

refi93: isCompatible is meant as a hard threshold beyond which the library would outright reject to interact with ledger having a cardano app older than that (ensured by `ensureLedgerAppVersionCompatible()` call at the beginning of all interactions except getVersion. Apps over that threshold would work as far as features supported by them go. Recommended version on the other hand is (usually) the latest available (or soon to be available) version of the ledger app which us (in the error messages) or the integrating party can ask the user to install in order to have access to all the features.

refi93: fair point though that the comments do not really help shed light on this, worth updating

src/types/public.ts
```diff
@@ -674,6 +684,33 @@ export type Witness = {
...
+ * @category Basic types
+ * @see [[TxAuxiliaryDataSupplementType]]
+ * @see [[SignedTransactionData]]
+ */
+export type TxAuxiliaryDataSupplement = {
+    type: TxAuxiliaryDataSupplementType.CATALYST_REGISTRATION,
+    /** Hash of the auxiliary data including the Catalyst registration */
+    auxiliaryDataHashHex: string,
+    /** Signature of the Catalyst registration payload by the staking key that was supplied */
+    signatureHex: string,
```
janmazak, line 711: Let's use a name specific to catalyst.

src/interactions/signTx.ts
```diff
@@ -334,8 +421,24 @@ function generateWitnessPaths(request: ParsedSigningRequest): ValidBIP32Path[] {
   }
 }
 
+function ensureRequestSupportedByAppVersion(version: Version, request: ParsedSigningRequest): void {
+  const auxiliaryData = request?.tx?.auxiliaryData;
+  const hasCatalystRegistration = auxiliaryData?.type == TxAuxiliaryDataType.TUPLE && auxiliaryData.metadata.type == TxMetadataType.CATALYST_REGISTRATION;
```
janmazak, line 426: It would probably be cleaner (though not much useful) to split support for TUPLE aux data and for Catalyst into separate features. Then we will not need this elaborate condition (brackets would help, though) and the flow here would naturally follow the enums describing aux data. Also, it will more clearly show the distinction between the two possible ways of serialization of aux data (dict / array).

src/types/internal.ts
```diff
@@ -55,9 +55,24 @@ export type ParsedNetwork = {
     networkId: Uint8_t
 }
 
+export const CATALYST_VOTING_PUBLIC_KEY_LENGTH = 32;
+
+export type CatalystVotingPublicKey = FixlenHexString<typeof CATALYST_VOTING_PUBLIC_KEY_LENGTH>
+
 export type  ParsedTxAuxiliaryData = {
     type: TxAuxiliaryDataType.ARBITRARY_HASH
     hashHex: FixlenHexString<32>
```
janmazak, line 64: Can this constant 32 be named?

src/types/public.ts
```diff
@@ -723,6 +768,63 @@ export type TxAuxiliaryDataArbitraryHashParams = {
...
+}
+
+/**
+ * Metadata entry to be included within the auxiliary data of the transaction
+ * @see [[TxMetadataType]]
+ * @see [[TxAuxiliaryData]]
+ */
+export type TxMetadata = {
+    type: TxMetadataType.CATALYST_REGISTRATION,
+    params: CatalystRegistrationParams
```
janmazak, line 798: Is 'params' specific to catalyst? If not, its type should not refer to catalyst; if yes, then perhaps it should be reflected in the name.

src/types/public.ts
```diff
@@ -723,6 +768,63 @@ export type TxAuxiliaryDataArbitraryHashParams = {
...
+export type CatalystRegistrationParams = {
+    /**
+     * Voting key to be registered given in hex
+     */
+    votingPublicKeyHex: string,
+
+    /**
+     * Path to the staking key to which voting rights would be associated
+     */
+    stakingPath: BIP32Path
```
janmazak, line 815: I'm surprised  by lack of commas here. Is it even allowed?

src/interactions/serialization/txAuxiliaryData.ts
```diff
@@ -0,0 +1,38 @@
+import type { ParsedTxAuxiliaryData, Uint8_t } from "../../types/internal";
+import { TxAuxiliaryDataType, TxMetadataType } from "../../types/internal";
+import { unreachable } from "../../utils/assert";
+import { hex_to_buf, uint8_to_buf } from "../../utils/serialize";
+
+export function serializeTxAuxiliaryData(
```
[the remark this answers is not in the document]
refi93: The rationale behind the client-facing "tuple" type was to explicitly acknowledge how the auxiliary data should be serialized, but giving it a second thought, I can see it that also as kind of a "premature optimization" in terms of a design we may never support (and if we would, this is probably the least complicated of the changes we'd face). It can also be misleading, suggesting that you can indeed pass other data along the catalyst registration which just isn't true.

So I think the most sensible/transparent for now is to actually have a dedicated Catalyst Registration type of auxiliary data without this apparently superfluous nesting. The way of serializing should be documented properly in the first place, and if we were to introduce another way of serializing/auxiliary scripts within auxiliary data, let's change the user-facing api then.

Seems like Trezor devs have similar thoughts about this design in Trezor: https://github.com/trezor/trezor-firmware/pull/1564/files#r616642829

src/interactions/serialization/catalystRegistration.ts
```diff
@@ -0,0 +1,27 @@
+import type { CatalystVotingPublicKey, ParsedAddressParams, Uint64_str,ValidBIP32Path } from "../../types/internal";
+import { hex_to_buf, path_to_buf, uint64_to_buf } from "../../utils/serialize";
+import { serializeAddressParams } from "./addressParams";
+
+export function serializeCatalystRegistrationVotingKey(votingPublicKey: CatalystVotingPublicKey): Buffer {
+    return Buffer.concat([
```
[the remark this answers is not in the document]
refi93: indeed, we are doing it like this consistently in other parts of the codebase (check e.g. serialization/txOther.ts) - it's maybe a bit superfluous but I can see value in the consistency regardless of whether you are passing one or multiple parameters, so I'd rather keep it like this.

src/interactions/serialization/txOther.ts
```diff
@@ -35,14 +35,6 @@ export function serializeTxTtl(
...
-    auxiliaryDataHashHex: HexString
-) {
-    return Buffer.concat([
-        hex_to_buf(auxiliaryDataHashHex)
-    ])
-}
-
 export function serializeTxValidityStart(
     validityIntervalStart: Uint64_str
 ) {
```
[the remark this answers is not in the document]
refi93: Other functions in the file don't have it either - given that those are one-liners and typescript infers the type it's imho acceptable

src/types/public.ts
refi93: interesting - worth enforcing consistency with a linting rule - I'd do it here: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/98 , for now I added commas just in types concerning this PR

src/interactions/signTx.ts
refi93: fixed by merging auxiliary data and metadata into a single type

src/types/public.ts
refi93: done - not sure why it doesn't seem to apply to type exports, but the rest is enforced which is still sharply better than the current state

src/types/public.ts
refi93: It's specific for Catalyst. We consistently use "params" to wrap the parameters of other typed objects such as Certificates, Pool Relays etc., so I'd rather not deviate from it unless a coherent rule/practice to follow can be established, so we don't end up with an inconsistent interface

src/interactions/serialization/txAuxiliaryData.ts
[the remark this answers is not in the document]
refi93: as discussed offline, I merged the TxAuxiliaryDataType and TxMetadataType into a single TxAuxiliaryDataType

src/interactions/serialization/catalystRegistration.ts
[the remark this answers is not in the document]
janmazak: OK, let's keep it like this, and change also
`return uint8_to_buf(Optional.NONE as Uint8_t)` in `serializePoolRelay` in `poolRegistrationCertificate.ts`

src/interactions/getVersion.ts
refi93: updated docstring to be clearer

src/interactions/serialization/txAuxiliaryData.ts
```diff
@@ -0,0 +1,38 @@
...
+        case TxAuxiliaryDataType.ARBITRARY_HASH: {
+            return Buffer.concat([
+                uint8_to_buf(auxiliaryDataTypesEncoding[auxiliaryData.type] as Uint8_t),
+                hex_to_buf(auxiliaryData.hashHex)
+            ]);
+        }
+        case TxAuxiliaryDataType.TUPLE: {
+            switch (auxiliaryData.metadata.type) {
+                case TxMetadataType.CATALYST_REGISTRATION:
+                    return Buffer.concat([
```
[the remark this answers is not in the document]
refi93: solving via https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/99#discussion_r617594019

src/interactions/serialization/txAuxiliaryData.ts
```diff
@@ -0,0 +1,38 @@
+import type { ParsedTxAuxiliaryData, Uint8_t } from "../../types/internal";
+import { TxAuxiliaryDataType, TxMetadataType } from "../../types/internal";
+import { unreachable } from "../../utils/assert";
+import { hex_to_buf, uint8_to_buf } from "../../utils/serialize";
+
+export function serializeTxAuxiliaryData(
+    auxiliaryData: ParsedTxAuxiliaryData,
+) {
```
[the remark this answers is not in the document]
refi93: solving via https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/99#discussion_r617594019

src/interactions/serialization/txOther.ts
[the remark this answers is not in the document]
janmazak: Well, I'm fine with not having it there; it's just that anyone new coming to the codebase will wonder why the return value is missing when we generally mention it explicitly.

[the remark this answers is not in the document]
refi93: of course, it may make sense to enforce it - if we do, it would make sense to set up a linter rule for that: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md , fix it globally, and going forward it would be sustained without any ambiguity

janmazak (approved):

src/parsing/address.ts
```diff
@@ -111,12 +119,15 @@ export function parseAddress(
         case AddressType.REWARD: {
             validate(params.stakingBlockchainPointer == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
             validate(params.stakingKeyHashHex == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
-            validate(params.stakingPath == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+            validate(params.spendingPath == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+
+            const stakingPath = parseBIP32Path(params.stakingPath!, InvalidDataReason.ADDRESS_INVALID_STAKING_KEY_PATH);
```
janmazak, line 124: I'm not sure about this `!` operator: as far as I understand, it does not do an actual check for `null`. 
Why is `stakingPath` treated differently than `spendingPath` in the above cases?

refi93: good point, it's not actually needed as the parseBIP32Path would throw if the staking key path is not defined

PR #102, opened 2021-04-21
Add flowgen

refi93:
Cloned PR from CF repo: https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/pull/20 with added integration to our CI pipeline so we know that the autogenerated types work properly and if anything goes wrong, can act on it


ppershing (approved):
Looks ok

PR #105, opened 2021-05-13
Enforce canonical ordering of assets in the tokenBundle (resolves #95)

### LINKED ISSUE #95: Enforce canonical ordering of assets in the tokenBundle and withdrawals
Currently Ledger serializes the assets/withdrawals in whatever order they are passed from outside - this however is a source of tx hash mismatches which are then hard to debug as the tx serialization logic of the wallets usually does so in canonical order. To at least make the debugging easier, we should throw an exception if the assets are not ordered in a canonical way

In case of multiasset it's also undesired that Ledger can receive the same policyId or assetName within the policyId twice and Ledger would "happily" serialize it into the map - such ill-formed map resuls in undefined behavior (will both entries take effect, or only one?) so it's beneficial to avoid such scenario altogether. Here's the related issue for the ledger app specifying the work needed on that end: https://github.com/vacuumlabs/ledger-app-cardano-shelley/issues/62

As part of the resolution of this issue, we should also submit a CIP establishing the way of serializing maps in (hw) wallets for the sake of clarity and interoperability between different wallets/hw wallet integrations

Solution proposed:
* verify valid structure of the [AssetGroup](https://vacuumlabs.github.io/ledgerjs-cardano-shelley/3.0.0/index.html#assetgroup) - i.e. that the policyIds are ordered alphabetically (they have the same length) and that assetNames within the respective policyIds are ordered by length first and then alphabetically (The canonical cbor specs specify that on the serialized representation of the keys, however, given they are of the same type, the prefix appended by CBOR doesn't interfere with the rules specified above - fell free to double-check, however)

Resources:
* Cardano CDDL (low-level tx format specs) on multiassets: https://github.com/input-output-hk/cardano-ledger-specs/blob/9ac75adb5597867a2318ae4628e559f7c4a9b2e4/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L250 (you can notice it's serialized as a map)
* RFC specs for CBOR elaborating on "canonical" serialization of maps: https://tools.ietf.org/html/rfc7049#section-3.9

balintklement:
Add checks to `signTransaction` to assure that policyIds within a tokenBundle and tokens within a policy are unique and ordered canonically.

### DIFF
--- a/src/errors/invalidDataReason.ts
+++ b/src/errors/invalidDataReason.ts
@@ -19,14 +19,16 @@ export enum InvalidDataReason {
   OUTPUTS_NOT_ARRAY = "outputs not an array",
 
   OUTPUT_INVALID_AMOUNT = "invalid amount in an output",
-  OUTPUT_INVALID_TOKEN_BUNDLE = "invalid multiasset token bundle in an output",
-  OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE = "multiasset token bundle too large",
-  OUTPUT_INVALID_TOKEN_POLICY = "invalid multiasset token policy",
-  OUTPUT_INVALID_ASSET_NAME =
-  "invalid asset name in the token bundle in an output",
-  OUTPUT_INVALID_ASSET_GROUP_TOKENS_NOT_ARRAY = "assetGroup tokens not an array",
-  OUTPUT_INVALID_ASSET_GROUP_TOKENS_TOO_LARGE = "assetGroup tokens too large",
-
+  OUTPUT_INVALID_POLICY_NAME = "invalid policy id in the token bundle in an output ",
+  OUTPUT_INVALID_TOKEN_BUNDLE_NOT_ARRAY = "invalid token bundle in an output - asset groups not an array",
+  OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE = "invalid token bundle in an output - too many asset groups",
+  OUTPUT_INVALID_TOKEN_BUNDLE_ORDERING = "invalid token bundle in an output - incorrect ordering of asset groups",
+  OUTPUT_INVALID_TOKEN_BUNDLE_NOT_UNIQUE = "invalid token bundle in an output - policyIds not unique",
+  OUTPUT_INVALID_ASSET_NAME = "invalid asset name in the token bundle in an output",
+  OUTPUT_INVALID_ASSET_GROUP_NOT_ARRAY = "invalid asset group in the token bundle in an output - tokens not an array",
+  OUTPUT_INVALID_ASSET_GROUP_TOO_LARGE = "invalid asset group in the token bundle in an output - too many tokens",
+  OUTPUT_INVALID_ASSET_GROUP_ORDERING = "invalid asset group in the token bundle in an output - incorrect ordering of tokens",
+  OUTPUT_INVALID_ASSET_GROUP_NOT_UNIQUE = "invalid asset group in the token bundle in an output - token names not unique",
 
   OUTPUT_INVALID_ADDRESS = "invalid address in an output",
 
--- a/src/parsing/transaction.ts
+++ b/src/parsing/transaction.ts
@@ -55,13 +55,35 @@ function parseToken(token: Token): ParsedToken {
 }
 
 function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
-    validate(isArray(assetGroup.tokens), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOKENS_NOT_ARRAY);
-    validate(assetGroup.tokens.length <= TOKENS_IN_GROUP_MAX, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOKENS_TOO_LARGE);
+    validate(isArray(assetGroup.tokens), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_NOT_ARRAY);
+    validate(assetGroup.tokens.length <= TOKENS_IN_GROUP_MAX, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOO_LARGE);
 
-    return {
-        policyIdHex: parseHexStringOfLength(assetGroup.policyIdHex, TOKEN_POLICY_LENGTH, InvalidDataReason.OUTPUT_INVALID_TOKEN_POLICY),
+    const parsedAssetGroup = {
+        policyIdHex: parseHexStringOfLength(assetGroup.policyIdHex, TOKEN_POLICY_LENGTH, InvalidDataReason.OUTPUT_INVALID_POLICY_NAME),
         tokens: assetGroup.tokens.map(t => parseToken(t))
     }
+
+    const assetNamesHex = parsedAssetGroup.tokens.map(t => t.assetNameHex)
+    const sortedAssetNames = [...assetNamesHex].sort( (n1, n2) => {
+        if (n1.length == n2.length) return n1.localeCompare(n2)
+        else return n1.length - n2.length
+    })
+    validate(JSON.stringify(assetNamesHex) == JSON.stringify(sortedAssetNames), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
+    validate(assetNamesHex.length == new Set(assetNamesHex).size, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_NOT_UNIQUE)
+
+    return parsedAssetGroup
+}
+
+
+function parseTokenBundle(tokenBundle: AssetGroup[]): ParsedAssetGroup[] {
+    const parsedTokenBundle = tokenBundle.map(ag => parseAssetGroup(ag))
+
+    const policyIds = parsedTokenBundle.map(ag => ag.policyIdHex)
+    const sortedPolicyIds = [...policyIds].sort()
+    validate(JSON.stringify(policyIds) == JSON.stringify(sortedPolicyIds), InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_ORDERING)
+    validate(policyIds.length == new Set(policyIds).size, InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_NOT_UNIQUE)
+
+    return parsedTokenBundle
 }
 
 export function parseTransaction(tx: Transaction): ParsedTransaction {
@@ -163,9 +185,9 @@ function parseTxOutput(
 ): ParsedOutput {
     const amount = parseUint64_str(output.amount, { max: MAX_LOVELACE_SUPPLY_STR }, InvalidDataReason.OUTPUT_INVALID_AMOUNT)
 
-    validate(isArray(output.tokenBundle ?? []), InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE);
+    validate(isArray(output.tokenBundle ?? []), InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_NOT_ARRAY);
     validate((output.tokenBundle ?? []).length <= ASSET_GROUPS_MAX, InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE);
-    const tokenBundle = (output.tokenBundle ?? []).map((ag) => parseAssetGroup(ag))
+    const tokenBundle = parseTokenBundle(output.tokenBundle ?? [])
 
     const destination = parseTxDestination(network, output.destination)
     return {
--- a/test/integration/__fixtures__/signTx.ts
+++ b/test/integration/__fixtures__/signTx.ts
@@ -1,5 +1,5 @@
 import type { DeviceOwnedAddress, SignedTransactionData, Transaction, TxInput, TxOutput, TxOutputDestination } from "../../../src/Ada";
-import { TxAuxiliaryDataSupplementType } from "../../../src/Ada";
+import {InvalidDataReason, TxAuxiliaryDataSupplementType} from "../../../src/Ada";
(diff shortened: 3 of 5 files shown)


src/parsing/transaction.ts
```diff
@@ -164,6 +164,25 @@ function parseTxOutput(
     validate((output.tokenBundle ?? []).length <= ASSET_GROUPS_MAX, InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE);
     const tokenBundle = (output.tokenBundle ?? []).map((ag) => parseAssetGroup(ag))
 
+    // validate that policyIDs and assetNames within the same policyIds are unique and sorted alphabetically
+    let prevPolicyIdHex = ""
+    for (let assetGroup of tokenBundle) {
```
refi93, line 169: I'm thinking if it wouldn't be more concise straightforward here to just sort the policies and compare them, same for the tokens, the solution could look like this:

```
const policyIds = tokenBundle.map((t) => t.policyIdHex);
const sortedPolicyIds = [...policyIds].sort()
if (JSON.stringify(policyIds) != JSON.stringify(sortedPolicyIds)) {
  InvalidDataReason.OUTPUT_INVALID_TOKEN_POLICY_ORDERING)
}
```
The same can be done for the individual token policies, but filling in a custom sorting function.

The advantage would be that it would be obvious from the code how the respective policies/asset names should be sorted to whoever would inspect it (so they can do it the same themselves) and you save yourself the hassle of keeping a `prevAssetNameHex` and comparing it - this technique makes sense in Ledger where the memory is limited, but in js we don't have to sacrifice simplicity for memory consumption.

refi93: ah, I forgot to check for duplicate elements as well in the solution above - arguably that could/should be another check and error message stating that. The check could consist in transforming the `policyIds` array of strings to a set and checking that its size matches the length of the array

src/parsing/transaction.ts
```diff
@@ -164,6 +164,25 @@ function parseTxOutput(
     validate((output.tokenBundle ?? []).length <= ASSET_GROUPS_MAX, InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE);
     const tokenBundle = (output.tokenBundle ?? []).map((ag) => parseAssetGroup(ag))
```
refi93, line 165: I'd consider moving the validation of ordering to a dedicated `parseTokenBundle` function for policyIds and similar validation could be done within the existing `parseAssetGroup()` function for the individual asset groups, to decouple the validation logic. Both are lower-level concerns than parsing of the overall tx output after all

src/parsing/transaction.ts
balintklement: you are right, this makes much more sense from a readability perspective!

src/parsing/transaction.ts
balintklement: This also makes sense to me.

src/parsing/transaction.ts
```diff
@@ -55,12 +55,27 @@ function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
     validate(isArray(assetGroup.tokens), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOKENS_NOT_ARRAY);
     validate(assetGroup.tokens.length <= TOKENS_IN_GROUP_MAX, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOKENS_TOO_LARGE);
 
+    const assetNames = assetGroup.tokens.map((t) => t.assetNameHex)
+    const sortedAssetNames = [...assetNames].sort()
```
refi93, line 59: this wouldn't work as intended - canonical CBOR ordering requires ordering by length first, then alphabetically

```javascript
["b", "aaaa", ].sort()
// yields ["aaaa", "b"], but canonical ordering is ["b", "aaaa"]
```

refi93: also, I would rather do this validation once the entries of the assetGroup are processed by the lower-level parsing functions, otherwise these checks may fail with weird errors, as the `policyIdHex` may for example be something that isn't even JSON-serializable

balintklement: damn, that would've been a stupid error. should I add some more tests to cover such corner-cases, or are the few i've made enough?

src/parsing/transaction.ts
```diff
@@ -55,12 +55,30 @@ function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
     validate(isArray(assetGroup.tokens), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOKENS_NOT_ARRAY);
     validate(assetGroup.tokens.length <= TOKENS_IN_GROUP_MAX, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_TOKENS_TOO_LARGE);
 
+    const assetNames = assetGroup.tokens.map((t) => t.assetNameHex)
+    const sortedAssetNames = [...assetNames].sort( (n1, n2) => {
+        if (n1.length == n2.length) return n1.localeCompare(n2)
```
refi93, line 60: as I wrote earlier, this would fail with generic errors if the integrating party happens to pass invalid data like null, or numbers or whatever else, so the integrating party would have a hard time telling whether the error is a result of some bug on their or even our side as we would not tell them that they passed an invalid policyId but it would just throw a cryptic error like "cannot find property length of ..." 

Therefore it seems sensible to parse the structure first, ensuring that the entries fullflill our assumptions about the keys being (hex)strings and only then proceed to sort/compare them.

The same observation applies for parseTokenBundle

src/parsing/transaction.ts
refi93: I agree - there's no harm in adding one more test that is supposed to fail if the validation function didn't sort based on length. A unit test would be ideal, but sadly we don't have such in the codebase (which is probably worth adding as part of some "housekeeping", but not now - feel free to create a follow up task for that)

src/parsing/transaction.ts
refi93: using localecompare for what is supposed to be a hex string seems like a bit of an overkill and a possibly confusing to a reader which may be misled into thinking that the assetNames could be generic strings (but they aren't)

refi93: ah I see, there's no native "non-locale" compare method for strings :/

balintklement: sorry, i somehow missed your previous comment on parsing before ordering validation.

src/parsing/transaction.ts
```diff
@@ -55,10 +55,32 @@ function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
...
         tokens: assetGroup.tokens.map(t => parseToken(t))
     }
+
+    const assetNameHexes = parsedAssetGroup.tokens.map(t => t.assetNameHex)
+    const sortedAssetNames = [...assetNameHexes].sort( (n1, n2) => {
+        if (n1.length == n2.length) return n1.localeCompare(n2)
+        else return n1.length - n2.length
+    })
+    validate(JSON.stringify(assetNameHexes) == JSON.stringify(sortedAssetNames), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
+    validate(assetNameHexes.length == new Set(assetNameHexes).size, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
```
refi93, line 69: nitpick - `assetNamesHex` would sound a bit more natural

test/integration/__fixtures__/signTx.ts
```diff
@@ -865,12 +946,12 @@ export const testsMary: TestcaseMary[] = [
     txBody: "a500818258203b40265111d8bb3c3c608d95b3a0bf83461ace32d79336579a1939b3aad1c0b700018282583901eb0baa5e570cffbe2934db29df0b6a3d7c0430ee65d4c3a7ab2fefb91bc428e4720702ebd5dab4fb175324c192dc9bb76cc5da956e3c8dff821904d2a2581c95a292ffee938be03e9bae5657982a74e9014eb4960108c9e23a5b39a24874652474436f696e1a00783862401904d2581c75a292ffee938be03e9bae5657982a74e9014eb4960108c9e23a5b39a1487564247542686911182f8258390114c16d7f43243bd81478e68b9db53a8528fd4fb1078d58d54a7f11241d227aefa4b773149170885aadba30aab3127cc611ddbc4999def61c1a006ca79302182a030a08182f",
```
refi93, line 946: this txBody should be updated as well - these are not returned from the device, they are kept here for reference and we dump them using the ledger app built in debug mode using the usb tool: https://ledger.readthedocs.io/en/latest/userspace/debugging.html#console-printing

Once you have the log file from the usb tool ftrom Ledger's execution for that specific fixture, you can extract the transaction body slices (dumped by the function computing the rolling tx hash) and merge them into a single transaction like this:

```javascript
const x = `<content of the log file>`
console.log(x.split('\n').filter((x) => x.includes('blake2b_256_append')).map((x) => x.split(' ')[1]).join(''))
```

test/integration/__fixtures__/signTx.ts
```diff
@@ -290,6 +308,69 @@ export const outputs: Record<
...
+        tokens: [
+          {
+            assetNameHex: "7564247542686911",
+            amount: "47",
+          },
+        ],
+      },
+    ],
+  },
+  multiassetInvalidTokenOrdering1: {
```
refi93, line 335: I'd choose some more descriptive name like `multiassetInvalidTokenOrderingSameLength` and `multiassetInvalidTokenOrderingDifferentLengths` for the fixtures, so it's clear at first sight how they differ

src/parsing/transaction.ts
```diff
@@ -55,10 +55,32 @@ function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
...
         tokens: assetGroup.tokens.map(t => parseToken(t))
     }
+
+    const assetNamesHex = parsedAssetGroup.tokens.map(t => t.assetNameHex)
+    const sortedAssetNames = [...assetNamesHex].sort( (n1, n2) => {
+        if (n1.length == n2.length) return n1.localeCompare(n2)
+        else return n1.length - n2.length
+    })
+    validate(JSON.stringify(assetNamesHex) == JSON.stringify(sortedAssetNames), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
+    validate(assetNamesHex.length == new Set(assetNamesHex).size, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
```
refi93, line 69: It's IMHO a bit misleading to state that the ordering is invalid in this case, this is rather a matter of the asset names being repeated so I'd rather create a separate error message for that case, same for policyIds below

test/integration/__fixtures__/signTx.ts
```diff
@@ -1030,3 +1111,37 @@ export const testsCatalystRegistration: TestcaseMary[] = [
...
+  {
+    testname: "Should reject transaction where tokens within an asset group are not ordered - length",
+    tx: {
+      ...maryBase,
+      outputs: [outputs.multiassetInvalidTokenOrdering2],
+    },
+    rejectReason: InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING,
+  },
+
+]
```
refi93, line 1147: I think other files have a newline at the end, so I'd add it here as well

refi93 (approved):
last few nitpicks, other than that, LGTM

refi93:
Also, don't forget to update the docstrings for TokenBundle (https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/0722de3d8f9ba690f1acbb5f330d1ea1e4f94ebc/src/types/public.ts#L279) types and AssetGroup (https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/0722de3d8f9ba690f1acbb5f330d1ea1e4f94ebc/src/types/public.ts#L262) , clarifying that they should be ordered canonically respective to the transaction CBOR format, in order to be valid, referencing the CBOR RFC: https://datatracker.ietf.org/doc/html/rfc7049#section-3.9

test/integration/__fixtures__/signTx.ts
```diff
@@ -1030,3 +1111,37 @@ export const testsCatalystRegistration: TestcaseMary[] = [
     },
   }
 ]
+
+export type InvalidTransactionTestcase = {
```
refi93, line 1115: this name is too generic, the transactions share in common that the tokenBundle is invalidly ordered so perhaps "InvalidTokenBundleOrderingTestcase" would be more suitable

janmazak (approved):
Squash it into a single commit, please, and rebase on top of `operatorapp-ts`. (We will release it together.)

test/integration/__fixtures__/signTx.ts
janmazak: I've made some small changes to tracing hashes, one can (should?) filter the logs for 
`blake2b_256_append_buffer_tx_body` and `blake2b_256_append_cbor_tx_body` to get serialized tx body. (There are similarly-named functions for aux data.)

@balintklement Are you going to write the documentation file, or should someone else do it?

src/errors/invalidDataReason.ts
```diff
@@ -22,10 +22,12 @@ export enum InvalidDataReason {
   OUTPUT_INVALID_TOKEN_BUNDLE = "invalid multiasset token bundle in an output",
   OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE = "multiasset token bundle too large",
   OUTPUT_INVALID_TOKEN_POLICY = "invalid multiasset token policy",
+  OUTPUT_INVALID_TOKEN_POLICY_ORDERING = "invalid multiasset token policy ordering",
```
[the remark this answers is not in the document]
balintklement: valid point, i changed this already to more expressing names

test/integration/__fixtures__/signTx.ts
balintklement: I will write the documentation file.

gabrielKerekes (approved):

janmazak (approved):

PR #107, opened 2021-05-18
Duplicate of #103 as I wasn't able to reopen it

gabrielKerekes:
This is a duplicate of #103 as it got closed when deleting old branches and I couldn't figure out a way to reopen it 😞 

All the changes requested in #103 should be addressed in https://github.com/vacuumlabs/ledgerjs-cardano-shelley/commit/c371359188a5fe0279ed42d39dc7b2c6fc662f73 or in the comments in the old PR.

### DIFF
--- a/src/errors/invalidDataReason.ts
+++ b/src/errors/invalidDataReason.ts
@@ -43,14 +43,12 @@ export enum InvalidDataReason {
 
   CERTIFICATES_NOT_ARRAY = "certificates not an array",
 
-  CERTIFICATE_MISSING_PATH = "path is required for one of the certificates",
-  CERTIFICATE_MISSING_POOL_KEY_HASH = "pool key hash missing in a certificate",
+  CERTIFICATE_INVALID_PATH = "one of the certificates contains an invalid path",
+  CERTIFICATE_INVALID_POOL_KEY_HASH = "one of the certificates contains an invalid pool key hash",
   CERTIFICATE_SUPERFLUOUS_POOL_KEY_HASH =
   "superfluous pool key hash in a certificate",
   CERTIFICATE_INVALID_TYPE = "invalid certificate type",
 
-  POOL_REGISTRATION_INVALID_POOL_KEY_HASH =
-  "invalid pool key hash in a pool registration certificate",
   POOL_REGISTRATION_INVALID_VRF_KEY_HASH =
   "invalid vrf key hash in a pool registration certificate",
   POOL_REGISTRATION_INVALID_PLEDGE =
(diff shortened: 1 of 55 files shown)


src/errors/invalidDataReason.ts
```diff
@@ -43,8 +43,8 @@ export enum InvalidDataReason {
 
   CERTIFICATES_NOT_ARRAY = "certificates not an array",
 
-  CERTIFICATE_MISSING_PATH = "path is required for one of the certificates",
-  CERTIFICATE_MISSING_POOL_KEY_HASH = "pool key hash missing in a certificate",
+  CERTIFICATE_INVALID_PATH = "path is required for one of the certificates",
+  CERTIFICATE_INVALID_POOL_KEY_HASH = "pool key hash missing in a certificate",
```
janmazak, line 47: Are these messages accurate?

src/types/public.ts
```diff
@@ -1017,7 +1017,7 @@ export enum TransactionSigningMode {
      * The API witnesses
      * - all non-null [[TxInput.path]] on `inputs`
      * - all [[Withdrawal.path]] on `withdrawals`
-     * - all `path` properties on `certificates` for Stake registering/deregistering/delegating certificate.
+    * - all `path` properties on `certificates` for Stake registering/deregistering/delegating certificate.
```
janmazak, line 1020: Some mis-indentation here?

janmazak (commented):
Tests: I am quite sure we want to have at least one (positive) test for each type of certificate, including stake deregistration (and each type for other tx elements). These integration tests are the only thing testing APDU handling and state machines in Ledger C code, and are thus very valuable during refactoring. Would you please add any missing tests?

Regarding the change in Mary tests, it is minor, so let's keep them as they are (assuming they do not fail).

gabrielKerekes:
There are tests for all the certificates in https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/develop/test/integration/__fixtures__/signTx.ts#L567. The test being added here was for a combination of a pool retirement certificate and any other certificate in `test/integration/__fixtures__/signTxPoolRegistration.ts` so I think that should be ok, no?

src/errors/invalidDataReason.ts
gabrielKerekes: Updated.

src/types/public.ts
gabrielKerekes: Oops, updated.

janmazak (approved):
Let's rename `poolRegistrationTestcases` to `poolRegistrationOwnerTestcases`.

Have we checked that the tests run with Ledger app version both 2.4.0 and 2.3.1?

gabrielKerekes:
`poolRegistrationTestcases` renamed to `poolRegistrationOwnerTestcases` and I also added "as operator" to the operator test names.

I also tested with FW version 2.3.2 and the only failing tests are regarding pool operator, pool retirement, operational certificate and get public key with account 150 and with purpose 1853' (cold keys).

PR #108, opened 2021-05-18
run tests on speculos instead of hw wallet

balintklement:
find more info on speculos on (github)[https://github.com/LedgerHQ/speculos] and (ledger.com)[https://speculos.ledger.com/user/usage.html]


janmazak:
Outdated by #128

PR #109, opened 2021-05-24
Add support for minting / burning of coins via transactions

balintklement:


src/errors/invalidDataReason.ts
```diff
@@ -18,17 +18,17 @@ export enum InvalidDataReason {
...
+  OUTPUT_INVALID_TOKEN_BUNDLE_NOT_ARRAY = "invalid token bundle in an output (or mint field) - asset groups not an array",
+  OUTPUT_INVALID_TOKEN_BUNDLE_TOO_LARGE = "invalid token bundle in an output (or mint field) - too many asset groups",
+  OUTPUT_INVALID_TOKEN_BUNDLE_ORDERING = "invalid token bundle in an output (or mint field) - incorrect ordering of asset groups",
+  OUTPUT_INVALID_TOKEN_BUNDLE_NOT_UNIQUE = "invalid token bundle in an output (or mint field) - policyIds not unique",
+  OUTPUT_INVALID_ASSET_NAME = "invalid asset name in the token bundle in an output (or mint field)",
+  OUTPUT_INVALID_ASSET_GROUP_NOT_ARRAY = "invalid asset group in the token bundle in an output (or mint field) - tokens not an array",
+  OUTPUT_INVALID_ASSET_GROUP_TOO_LARGE = "invalid asset group in the token bundle in an output (or mint field) - too many tokens",
+  OUTPUT_INVALID_ASSET_GROUP_ORDERING = "invalid asset group in the token bundle in an output (or mint field) - incorrect ordering of tokens",
+  OUTPUT_INVALID_ASSET_GROUP_NOT_UNIQUE = "invalid asset group in the token bundle in an output (or mint field) - token names not unique",
```
balintklement, line 31: With this commit, the same function (`parseTokenBundle`) is now used internally for parsing both the output and mint field. With this function's current structure it is not trivial to determine, where a potential parsing error occurred. What do you think, is it okay to return a generic exception (maybe with another wording), or should the exception's origin be explicitly stated in the error message?

src/types/public.ts
```diff
@@ -725,6 +725,33 @@ export type Withdrawal = {
...
+    policyIdHex: string,
+    /**
+     * The keys must be sorted lowest value to highest to reflect a valid canonical CBOR.
+     * The sorting rules (as described in the [CBOR RFC](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9)) are:
+     *  * if two keys have different lengths, the shorter one sorts earlier;
+     *  * if two keys have the same length, the one with the lower value in lexical order sorts earlier.
+     */
+    tokens: Array<MintToken>,
+};
+
```
balintklement, line 754: The data type on the public interface is the same for the Tokens in the output and in the mint field. However, I added these novel types, as the documentation of these types differ. I think having a generic documentation for both tokenBundles can be confusing. What do you think about that?

src/types/public.ts
```diff
@@ -725,6 +725,33 @@ export type Withdrawal = {
     amount: bigint_like,
 };
 
+/**
+ * Describes a single token to be minted or burned during the transaction
+ * @category ~Mary~ --> todo?
```
balintklement, line 730: Which era/sub-era does this update belong to?

src/utils/parse.ts
```diff
@@ -64,6 +66,41 @@ export const isUintStr = (data: unknown, constraints: { min?: string, max?: stri
...
+export const isIntStr = (data: unknown, constraints: { min?: string, max?: string }): data is string => {
+    const min = constraints.min ?? MIN_INT_64_STR
+    const max = constraints.max ?? MAX_INT_64_STR
+
+    let hasValidFormat = isString(data)
+        // check format via RegExp
+        && /^-?[0-9]*$/.test(data)
+        // Length checks
+        && data.length > 0
+        && data.length <= Math.max(min.length, max.length)
```
janmazak, line 88: Doesn't this allow a positive integer one digit too long? (If checked later, perhaps a short explanatory comment would be handy.)

src/types/internal.ts
```diff
@@ -95,6 +99,7 @@ export type ParsedTransaction = {
     withdrawals: ParsedWithdrawal[]
     auxiliaryData: ParsedTxAuxiliaryData | null
     validityIntervalStart: Uint64_str | null
+    mintInstructions: Array<ParsedAssetGroup<Int64_str>> | null
```
janmazak, line 102: Perhaps just `mint`, it conforms with CDDL.

src/types/internal.ts
```diff
@@ -14,6 +14,10 @@ export type Uint32_t = number & { __type: 'uint32_t' }
 export type Uint16_t = number & { __type: 'uint16_t' }
 export type Uint8_t = number & { __type: 'uint8_t' }
 
+export type Int64_str = string & { __type: 'int64_t' }
```
janmazak, line 17: A slight reorganization: let's move `ValidBIP32Path` from above to the end of this type list since it's the most complex type, and then we can keep the integer types together.

src/parsing/transaction.ts
```diff
@@ -122,6 +125,12 @@ export function parseTransaction(tx: Transaction): ParsedTransaction {
         ? null
         : parseUint64_str(tx.validityIntervalStart, {}, InvalidDataReason.VALIDITY_INTERVAL_START_INVALID)
 
+    // mint instructions
+    validate(isArray(tx.mintInstructions ?? []), InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_NOT_ARRAY)
```
janmazak, line 129: I'd keep this error separate from outputs.

src/errors/invalidDataReason.ts
janmazak: I don't have a strong preference; it's nice to report the precise location of the error, but not that necessary here. I asked @refi93 for an opinion of how to best implement distinct errors.

One option would be to pass a map with errors (sound like an overkill for this case and we don't do anything like that yet), another to pass error location description as an addtitional string argument and just concatenate it with the generic error description (seems more suitable here and could be used in other similar cases, too).

src/parsing/transaction.ts
janmazak: Maybe this check could be made part of "parseTokenBundle"?

src/utils/parse.ts
balintklement: you are right, this would allow a positive integer one digit too long. but actually the string comparisons in `isValidPositiveNumber`/`isValidNegativeNumber` that assert the number is not over/under the maximum/minimum would catch too long strings anyways, or not?

src/errors/invalidDataReason.ts
janmazak: So I've discussed it with @refi93  and we think it is not worth it to have separate error messages; let's just rename the existing ones (no OUTPUT, perhaps replace it with MULTIASSET), and do not mention anything about output or mint field.

src/utils/parse.ts
janmazak: I think yes. It's just that it if a validation is incomplete, as in this case, it should be explicitly mentioned. Alternatively, we can drop the length check here and keep only the checks in sub-cases. That would perhaps be the cleanest solution.

src/parsing/transaction.ts
janmazak: Static compile-time validation is independent from runtime-validation AFAIK. That's why we check if strings are actually strings. As it is now, we only runtime-check that the argument passed to `parseTokenBundle` is an array. But there is no reason to split the validation into two parts (we only check whether the elements of the array are of proper type inside the function, despite the argument being declared as if they were already `AssetGroup`s). So all of the runtime checks can be done within `parseTokenBundle`, including the array check.

Compare it with validation of inputs: we are calling `tx.inputs.map()`, and that can only be done for an array, so we have to validate the array first.

src/utils/parse.ts
balintklement: another corner case came to my mind: do we want to accept `-0`?

balintklement: i guess not. if we already reject e.g. `052`, we should also reject `-0`

janmazak: I do not mind accepting negative zero; it is in somewhat distinct from positive (unsigned) zero in javascript, see e.g.
https://medium.com/coding-at-dawn/is-negative-zero-0-a-number-in-javascript-c62739f80114
(It makes the serialization format of integers ambiguous, but since we are already willing to accept string, number, BigInt as input, accepting negative zeros would be quite in line with this philosophy of not putting unnecessary obstacles in the way of ledgerjs users.)

balintklement: all right!

src/types/public.ts
gabrielKerekes: I think it shouldn't be a problem to have a single type for both outputs and for minting. I don't think it's worth the duplication. The docs can IMHO be easily written in a way to describe the purpose of the types.

src/parsing/transaction.ts
```diff
@@ -40,27 +40,30 @@ function parseCertificates(certificates: Array<Certificate>): Array<ParsedCertif
...
+
+function parseToken<Type>(token: Token, parseFn: ParseNumberFunction<Type>): ParsedToken<Type> {
     const assetNameHex = parseHexString(token.assetNameHex, InvalidDataReason.OUTPUT_INVALID_ASSET_NAME)
     validate(
         token.assetNameHex.length <= ASSET_NAME_LENGTH_MAX * 2,
         InvalidDataReason.OUTPUT_INVALID_ASSET_NAME
     )
 
-    const amount = parseUint64_str(token.amount, {}, InvalidDataReason.OUTPUT_INVALID_AMOUNT)
+    const amount = parseFn(token.amount, {}, InvalidDataReason.OUTPUT_INVALID_AMOUNT)
```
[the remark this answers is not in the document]
balintklement: I don't really like using boolean parameters. I initially wanted to decide depending on the type of `Type`, but I think there is no way to access this information during runtime, as the parameter `token.amount` has type `bigint_like`, not `Int64_str` or `Uint64_str`.

If my understanding is correct, the same also holds for a potential generic function `parseNumber64_str<T>(value: bigint_like): T`. You would need to pass some additional information - either a function, or an Enum (I like this slightly more than a boolean flag).

src/parsing/transaction.ts
```diff
@@ -71,20 +71,21 @@ function parseAssetGroup<Type>(assetGroup: AssetGroup, parseTokenAmountFunction:
         if (n1.length == n2.length) return n1.localeCompare(n2)
```
janmazak, line 71: Shouldn't we use `===` for comparisons? (More than one place.)

src/interactions/serialization/txInit.ts
```diff
@@ -37,6 +37,7 @@ export function serializeTxInit(tx: ParsedTransaction, signingMode: TransactionS
         _serializeOptionFlag(tx.ttl != null),
         _serializeOptionFlag(tx.auxiliaryData != null),
         _serializeOptionFlag(tx.validityIntervalStart != null),
+        _serializeOptionFlag(tx.mint != null),
```
balintklement, line 40: Added an option flag to indicate whether a mint field is present.
(@mkv-vcm)

src/interactions/signTx.ts
```diff
@@ -26,7 +26,8 @@ const enum P1 {
   STAGE_CERTIFICATES = 0x06,
   STAGE_WITHDRAWALS = 0x07,
   STAGE_VALIDITY_INTERVAL_START = 0x09,
-  STAGE_CONFIRM = 0x0a,
+  STAGE_MINT = 0x0a,
+  STAGE_CONFIRM = 0x0b,
```
balintklement, line 30: Is it okay to also change the confirm phase's code? I thought it might be important to keep the ordering of the steps. (@mkv-vcm)

src/interactions/signTx.ts
```diff
@@ -117,6 +99,34 @@ function* signTx_addOutput(
   })
 }
 
+export type SerializingFunction<Type> = (val: Type) => Buffer
+
+function* signTx_addTokenBundle<Type>(tokenBundle: ParsedAssetGroup<Type>[], p1: number, parseFn: SerializingFunction<Type>) {
+    const enum P2 {
+        ASSET_GROUP = 0x31,
+        TOKEN = 0x32,
+    }
```
balintklement, line 108: Is my understanding correct, that the `P2` variable _always_ is should be in the range 0x30...0x3f?
Or should this value be different when transmitting the output than when transmitting the mint?

src/utils/serialize.ts
```diff
@@ -53,6 +53,16 @@ export function uint64_to_buf(value: Uint64_str): Buffer {
...
+export function int64_to_buf(value: Int64_str): Buffer {
+    assert(isInt64str(value), 'invalid int64_str')
+
+    const data = bs10.decode(value)
+    assert(data.length <= 8, "excessive data")
+
+    const padding = Buffer.alloc(8 - data.length)
+    return Buffer.concat([padding, data])
+}
+
```
balintklement, line 65: the function should be the same for `int64` and `uint64` - should i better rename the original one and avoid the duplicate?

src/interactions/signTx.ts
```diff
@@ -433,6 +443,35 @@ function* signTx_setValidityIntervalStart(
   })
 }
 
+function* signTx_setMint(
+    mint: Array<ParsedAssetGroup<Int64_str>>
+): Interaction<void> {
+    const enum P2 {
+        BASIC_DATA = 0x30,
+        CONFIRM = 0x33,
+    }
```
balintklement, line 452: same question as above:
Is my understanding correct, that the P2 variable always is should be in the range 0x30...0x3f?
Or should this value be different when transmitting the output than when transmitting the mint?

src/parsing/transaction.ts
balintklement: i honestly was not aware of the exact difference - i will check this out.

src/interactions/signTx.ts
mkv-vcm: I don't think changing theses is a good idea. It would break compatibility with other ledger app versions. If you check the transaction_body in the cddl, every code here except the validity interval is two above the one in the cddl. So 0x0b (9+2) could be Mint. Or any other number, I'm not sure who else do we want to be compatible with. I don't think order is particularly important.

src/interactions/signTx.ts
mkv-vcm: I don't think reusing the values could be any problem. Minting is eerily similar to simple multiasset outputs, so the more we reuse the better.

src/parsing/transaction.ts
balintklement: I don't think it makes a difference in this case, as we have a (verified) string array, and calling the `length` function on a string always returns a number, so the types will be the same. however, it could be a reasonable paradigm to always use `===` for comparisons, regardless of whether it's necessary.

src/parsing/transaction.ts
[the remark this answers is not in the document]
gabrielKerekes: Oh, I didn't realise the runtime part :+1: Let's keep the function then. Maybe just `parseAmountFn` or `parseTokenAmountFn` might be better names.

src/interactions/signTx.ts
```diff
@@ -433,6 +443,35 @@ function* signTx_setValidityIntervalStart(
...
+
+    // Basic data
+    {
+        yield send({
+            p1: P1.STAGE_MINT,
+            p2: P2.BASIC_DATA,
+            data: serializeMintBasicParams(mint),
+            expectedResponseLength: 0,
+        })
+    }
```
balintklement, line 462: Similar to `signTx_addOutput` function, I put this block within extra curly braces. Is this necessary?

src/parsing/transaction.ts
janmazak: Definitely. Always use `===` unless you have a specific reason to use `==`.

src/interactions/signTx.ts
janmazak: Definitely better to avoid changing these. The codes are meaningless, they just should be distinct.

src/interactions/signTx.ts
janmazak: There is no specific range for P2. It's just that outputs were 3 in some enum, so we used 0x3* for those P2 values. I think it is reasonable to keep the same P2 values for minting.

src/interactions/serialization/txOther.ts
```diff
@@ -49,4 +50,25 @@ export function serializeTxWitnessRequest(
     return Buffer.concat([
         path_to_buf(path),
     ])
+}
+
+export function serializeAssetGroup<Type>(assetGroup: ParsedAssetGroup<Type>) {
```
janmazak, line 55: I think that T would be better than Type here. Shorter, if nothing else. Unless you want to give a descriptive name to the type, which is irrelevant here.

src/interactions/serialization/txOther.ts
```diff
@@ -49,4 +50,25 @@ export function serializeTxWitnessRequest(
...
+        hex_to_buf(assetGroup.policyIdHex),
+        uint32_to_buf(assetGroup.tokens.length as Uint32_t),
+    ])
+}
+
+export function serializeToken<Type>(token: ParsedToken<Type>, parseFn: SerializingFunction<Type>) {
+    return Buffer.concat([
+        uint32_to_buf(token.assetNameHex.length / 2 as Uint32_t),
+        hex_to_buf(token.assetNameHex),
+        parseFn(token.amount),
```
janmazak, line 66: `parseFn` sounds wrong since the token is already parsed?

src/interactions/signTx.ts
janmazak: If you keep the same values for outputs and mint, it might be useful later (simplify things in ledger app implementation).

src/interactions/signTx.ts
janmazak: It's not necessary. You can remove the braces in both places.

src/types/internal.ts
```diff
@@ -49,14 +51,14 @@ export type ParsedCertificate = {
 export const TOKEN_POLICY_LENGTH = 28
 
 
-export type ParsedToken = {
+export type ParsedToken<Type> = {
```
janmazak, line 54: Here, I would consider giving the type a descriptive name, e.g. IntegerType, and perhaps added an explanatory comment that it is used with both uint64 for outputs and int64 for minting.

src/types/public.ts
```diff
@@ -725,6 +725,33 @@ export type Withdrawal = {
     amount: bigint_like,
 };
 
+/**
+ * Describes a single token to be minted or burned during the transaction
+ * @category Mary
+ * @see [[MintAssetGroup]]
+ */
+export type MintToken = {
```
janmazak, line 733: Didn't we agree on avoid mint-specific types for tokens, and using the same type for both output and mint?

src/utils/serialize.ts
janmazak: I would keep the functions separate. Are we sure that `bs10.decode` actually works for negative numbers?

src/interactions/signTx.ts
mkv-vcm: It probably won't, since the handlers are typically implemented in separate files, with local enums for consts. But I would argue to keep the same for multiasset output and mint the same just for sake of consistency.

src/interactions/signTx.ts
balintklement: all right, then i will just switch them (confirm stays 0x0a, mint will be 0x0b)

src/interactions/serialization/txOther.ts
balintklement: you are right, I changed the name to `serializeTokenAmountFn`

src/types/public.ts
balintklement: yes, i just didn't push the update yet.

src/utils/serialize.ts
balintklement: after adjusting the alphabet in l.9 (adding the "-" sign) the basic unit test:
`bs10.encode(bs10.decode(nmb)) == nmb`
passes for both negative and positive numbers, and
`bs10.decode(nmb) != bs10.decode(-nmb)`.

However, changing the alphabet also changes the coding, so this alphabet change will also be necessary in _ledger-app_.

mkv-vcm: What does this mean? What is I.9, and won't this break compatibility with previous ledger versions?

balintklement: you are right, that would break compatibility. l.9 is line 9, where `bs10` is defined: `const bs10 = basex("0123456789")`.
i undid this change, and added a new base encoder `bs11` based on an extended alphabet, which includes `-`. this way we can use `bs11` for signed numbers, and keep backwards compatibility by keeping the decoding for unsigned number.

src/utils/serialize.ts
```diff
@@ -1,12 +1,13 @@
...
 
-import type { FixlenHexString, HexString, Uint8_t, Uint16_t, Uint32_t, Uint64_str } from "../types/internal"
+import type { FixlenHexString, HexString, Int64_str, Uint8_t, Uint16_t, Uint32_t, Uint64_str } from "../types/internal"
 import { assert } from './assert'
-import { isHexString, isUint8, isUint16, isUint32, isUint64str, isValidPath } from "./parse"
+import { isHexString, isInt64str, isUint8, isUint16, isUint32, isUint64str, isValidPath } from "./parse"
 
 // We use bs10 as an easy way to parse/encode amount strings
 const bs10 = basex("0123456789")
+const bs11 = basex("-0123456789")
```
janmazak, line 10: I guess I'd like to keep this local to the function using it. The name bs11 in itself could be very misleading (we are not doing base-11 decoding), could you come up with a better one? 

I don't understand how/why the basex library works, it does not make any sense whatsoever to me to include `-` as part of the alphabet --- well, not in the ordinary meaning for negative numbers; they use `-` as a part of base-66 alphabet in https://www.npmjs.com/package/base-x.

Do you understand why it works for our case as expected? Have you tried it extensively? I guess an explanation should be added in a comment; right now, it looks like some kind of hack one is scared of modifying because of introducing potential bugs (at least I am very wary about it).

src/types/internal.ts
```diff
@@ -95,6 +98,7 @@ export type ParsedTransaction = {
     withdrawals: ParsedWithdrawal[]
     auxiliaryData: ParsedTxAuxiliaryData | null
     validityIntervalStart: Uint64_str | null
+    mint: Array<ParsedAssetGroup<Int64_str>> | null
```
janmazak, line 101: Hm, we need to pass in witness paths. That part is not designed yet. Please add just a TODO comment here. I guess @gabrielKerekes might have a suggestion on the API?

janmazak (commented):
Apart from that bs11 part I don't understand, it looks OK.

Should not be merged since witnesses are not processed properly yet, and we have not tried it (no counterparty code ready yet).

Please add the label "do not merge" to this PR.

mkv-vcm (commented):
@refi93 I would like to get your opinion about this commit. The base-x approach was not working particularly well, especially for negative numbers, but even for unsigned it was not ideal (it produces shorter buffers than what is read on the other side, never ideal).

refi93:
@mkv-vcm sorry I got to see your comment just now - if you mean commit https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/109/commits/bd82646314b5ad7cb6b35b7595457b3299473edb then I'm fine with the approach to signed integers - makes sense to me to pull in a library to deal with negative numbers and this one seems to have no further dependencies

test/unit/parse.test.ts
```diff
@@ -1,63 +1,96 @@
...
 import {InvalidDataReason} from "../../src/errors"
 import {assert} from "../../src/utils/assert"
 import {parseInt64_str, parseUint64_str} from "../../src/utils/parse"
 
 describe('test parsing of integers', () => {
-    it('parse int64 correctly', () => {
-        let nmb = "1235543"
-        let result = parseInt64_str(nmb, {}, InvalidDataReason.INPUT_INVALID_TX_HASH)
-        expect(result).to.equal("1235543")
+    it("decode signed 1", () => {
```
gabrielKerekes, line 9: As the tests are fairly repetitive I think it would be nice have a list of test vectors which would get iterated. As is done in e.g. `signTx.test.ts` but the fixtures could be kept in this file.

Or instead of `decode signed 1` the name could contain the number being decoded i.e. `decode signed - -123456`, `decode signed - 9223372036854775807` etc.

test/integration/__fixtures__/signTx.ts
```diff
@@ -161,23 +161,82 @@ const destinations: Record<
...
+                },
+                {
+                    // fingerprint: asset17jd78wukhtrnmjh3fngzasxm8rck0l2r4hhyyt (and incremented)
+                    assetNameHex: "1e349c9bdea19fd6c147626a5260bc44b71635f398b67c59881df20b",
+                    amount: "-9223372036854775808",
+                },
+            ],
+        },
+    ],
+    mintCanonicalOrderingPolicy: [
```
gabrielKerekes, line 197: The invalid test objects could be named `mintInvalid...`

test/unit/parse.test.ts
mkv-vcm: I only did the primitive parsing tests, since the other three are not that numerous and I'm not even certain what they test, except the validation code. Might need to remove them, because really.

src/types/public.ts
```diff
@@ -995,7 +996,16 @@ export type Transaction = {
      * Validity start (block height) if any.
      * Transaction becomes valid only starting from this block height.
      */
-    validityIntervalStart?: bigint_like | null
+    validityIntervalStart?: bigint_like | null,
+    /**
+     * Mint or burn instructions (if any).
+     * Assets to be minted (token amount positive) or burned (token amount negative) with the transaction.
+     * If not null, the entries' keys (policyIds) must be sorted lowest value to highest to reflect a canonical CBOR.
+     * The sorting rules (as described in the [CBOR RFC](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9)) are:
```
janmazak, line 1004: We need to refer to the CIP not written yet (which is likely to conform to the RFC).

janmazak:
Will be merged via another PR.

PR #111, opened 2021-06-08
Temporarily remove asset groups canonical ordering validation

gabrielKerekes:
Validation is being removed because it is causing problems for cardano-hw-cli users. Also the exact canonical ordering is yet to be defined.

I've decided to just comment out the code so it'll be easier to add later.

### DIFF
--- a/src/parsing/transaction.ts
+++ b/src/parsing/transaction.ts
@@ -64,12 +64,14 @@ function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
     }
 
     const assetNamesHex = parsedAssetGroup.tokens.map(t => t.assetNameHex)
-    const sortedAssetNames = [...assetNamesHex].sort( (n1, n2) => {
-        if (n1.length == n2.length) return n1.localeCompare(n2)
-        else return n1.length - n2.length
-    })
-    validate(JSON.stringify(assetNamesHex) == JSON.stringify(sortedAssetNames), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
     validate(assetNamesHex.length == new Set(assetNamesHex).size, InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_NOT_UNIQUE)
+
+    // enforcing of asset order is removed for now and will be added back after the ordering is properly defined by a CIP
+    // const sortedAssetNames = [...assetNamesHex].sort( (n1, n2) => {
+    //     if (n1.length == n2.length) return n1.localeCompare(n2)
+    //     else return n1.length - n2.length
+    // })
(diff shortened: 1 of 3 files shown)


refi93:
> Hm, it seems the CI checks are failing. It also seems that is is normal for PRs? Do we know why that is?

fixed here: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/112

refi93 (approved):

src/parsing/transaction.ts
```diff
@@ -64,11 +64,14 @@ function parseAssetGroup(assetGroup: AssetGroup): ParsedAssetGroup {
...
-    })
-    validate(JSON.stringify(assetNamesHex) == JSON.stringify(sortedAssetNames), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
+    
+    // enforcing of asset order is removed for now and will be added back after the ordering is properly defined by a CIP
+    // const sortedAssetNames = [...assetNamesHex].sort( (n1, n2) => {
+    //     if (n1.length == n2.length) return n1.localeCompare(n2)
+    //     else return n1.length - n2.length
+    // })
+    // validate(JSON.stringify(assetNamesHex) == JSON.stringify(sortedAssetNames), InvalidDataReason.OUTPUT_INVALID_ASSET_GROUP_ORDERING)
+
```
janmazak, line 74: I guess I would first check if the asset names are unique (so just move these comments a few lines below).

src/parsing/transaction.ts
```diff
@@ -79,8 +82,11 @@ function parseTokenBundle(tokenBundle: AssetGroup[]): ParsedAssetGroup[] {
     const parsedTokenBundle = tokenBundle.map(ag => parseAssetGroup(ag))
 
     const policyIds = parsedTokenBundle.map(ag => ag.policyIdHex)
-    const sortedPolicyIds = [...policyIds].sort()
-    validate(JSON.stringify(policyIds) == JSON.stringify(sortedPolicyIds), InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_ORDERING)
+    
+    // enforcing of policies order is removed for now and will be added back after the ordering is properly defined by a CIP
+    // const sortedPolicyIds = [...policyIds].sort()
+    // validate(JSON.stringify(policyIds) == JSON.stringify(sortedPolicyIds), InvalidDataReason.OUTPUT_INVALID_TOKEN_BUNDLE_ORDERING)
+
```
janmazak, line 89: The same as above, first check for duplicity.

janmazak (approved):
I did not run anything, but it LGTM.

It should be tried with non-canonically-ordered tokens to make sure we did not miss something.

gabrielKerekes:
I tested it with non-canonically-ordered tokens once more just to be sure 😉  And I also changed the order of the commented code.

PR #112, opened 2021-06-08
Fix typedoc prepare script to use npm

refi93:
Reason: typedoc has package.lock, and yarn install (what was being previously used) would not install locked
dependencies but the newest ones, hence breaking on newer packages than the ones that actually work being pulled

Further context: we are using a custom fork of typedoc which doesn't sort the properties, which gets downloaded and built within our repo: https://github.com/ppershing/typedoc/commit/f32157ca3f0a0f64cda95ca3c622bbbfa59dfc14

I noticed the installation of the dependencies within the typedoc build step results in vulnerabilities being reported but these dependencies don't interfere with the dependencies of our package and is ran as a separate script to build the docs

Additionaly, yarn audit on ledgerjs repo itself was failing with glob-parent dependency so I fixed that as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   },
   "homepage": "https://github.com/cardano-foundation/ledgerjs-hw-app-cardano",
   "scripts": {
-    "prepare": "cd node_modules/typedoc && yarn install && yarn build",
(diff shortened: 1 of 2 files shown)


gabrielKerekes (approved):
🧙

PR #113, opened 2021-06-09
Update changelog, bump gh-pages package

refi93:


janmazak (approved):

refi93:
The last yarn audit error (normalize-url) seems to be not fixable without updating the gh-pages package we use to publish docs as it depends on a too old version of the package, so even if we force a newer one, it wouldn't work. I'd advise not fixing it right now and waiting for gh-pages to update itself: https://github.com/tschaub/gh-pages/issues/392

This is a dev dependency, so projects depending on ourl ibrary won't be exposed to gh-pages and its subdependencies

PR #115, opened 2021-06-21
[WIP] derive address

mkv-vcm:
The part about parsing/address.ts:parseAddress is the one that makes me uncertain. I have to rephrase ShelleyAddressParams but I don't exactly know how, it should have a union of spendingpath and spendingscripthash, and the whole addressParams function needs a serious rethink, because of the overlapping and criss-crossing requirements from the various address types.

### DIFF
--- a/src/interactions/serialization/addressParams.ts
+++ b/src/interactions/serialization/addressParams.ts
@@ -8,6 +8,7 @@ function serializeStakingChoice(stakingChoice: StakingChoice): Buffer {
         [StakingChoiceType.STAKING_KEY_PATH]: 0x22,
         [StakingChoiceType.STAKING_KEY_HASH]: 0x33,
         [StakingChoiceType.BLOCKCHAIN_POINTER]: 0x44,
+        [StakingChoiceType.STAKING_SCRIPT_HASH]: 0x55,
     } as const
 
     switch (stakingChoice.type) {
@@ -16,7 +17,8 @@ function serializeStakingChoice(stakingChoice: StakingChoice): Buffer {
             uint8_to_buf(stakingChoicesEncoding[stakingChoice.type] as Uint8_t),
         ])
     }
-    case StakingChoiceType.STAKING_KEY_HASH: {
+    case StakingChoiceType.STAKING_KEY_HASH:
+    case StakingChoiceType.STAKING_SCRIPT_HASH: {
         return Buffer.concat([
             uint8_to_buf(stakingChoicesEncoding[stakingChoice.type] as Uint8_t),
             hex_to_buf(stakingChoice.hashHex),
--- a/src/interactions/serialization/addressParams.ts
+++ b/src/interactions/serialization/addressParams.ts
@@ -1,47 +1,47 @@
-import { ParsedAddressParams, SpendingChoice, SpendingChoiceType, StakingChoice, Uint8_t } from "../../types/internal"
-import { AddressType, StakingChoiceType } from "../../types/internal"
+import { ParsedAddressParams, SpendingChoice, SpendingDataSourceType, StakingChoice, Uint8_t } from "../../types/internal"
+import { AddressType, StakingDataSource } from "../../types/internal"
 import { hex_to_buf, path_to_buf, uint8_to_buf, uint32_to_buf } from "../../utils/serialize"
 
 function serializeSpendingChoice(spendingChoice: SpendingChoice): Buffer {
     switch (spendingChoice.type) {
-    case SpendingChoiceType.PATH:
+    case SpendingDataSourceType.PATH:
         return path_to_buf(spendingChoice.path)
-    case SpendingChoiceType.SCRIPT_HASH:
+    case SpendingDataSourceType.SCRIPT_HASH:
         return hex_to_buf(spendingChoice.scriptHash)
-    case SpendingChoiceType.NONE:
+    case SpendingDataSourceType.NONE:
         return Buffer.alloc(0)
     }
 }
 
 function serializeStakingChoice(stakingChoice: StakingChoice): Buffer {
     const stakingChoicesEncoding = {
-        [StakingChoiceType.NO_STAKING]: 0x11,
-        [StakingChoiceType.STAKING_KEY_PATH]: 0x22,
-        [StakingChoiceType.STAKING_KEY_HASH]: 0x33,
-        [StakingChoiceType.BLOCKCHAIN_POINTER]: 0x44,
-        [StakingChoiceType.STAKING_SCRIPT_HASH]: 0x55,
+        [StakingDataSource.NONE]: 0x11,
+        [StakingDataSource.KEY_PATH]: 0x22,
+        [StakingDataSource.KEY_HASH]: 0x33,
+        [StakingDataSource.BLOCKCHAIN_POINTER]: 0x44,
+        [StakingDataSource.SCRIPT_HASH]: 0x55,
     } as const
 
     switch (stakingChoice.type) {
-    case StakingChoiceType.NO_STAKING: {
+    case StakingDataSource.NONE: {
         return Buffer.concat([
             uint8_to_buf(stakingChoicesEncoding[stakingChoice.type] as Uint8_t),
         ])
     }
-    case StakingChoiceType.STAKING_KEY_HASH:
-    case StakingChoiceType.STAKING_SCRIPT_HASH: {
+    case StakingDataSource.KEY_HASH:
+    case StakingDataSource.SCRIPT_HASH: {
         return Buffer.concat([
             uint8_to_buf(stakingChoicesEncoding[stakingChoice.type] as Uint8_t),
             hex_to_buf(stakingChoice.hashHex),
         ])
     }
-    case StakingChoiceType.STAKING_KEY_PATH: {
+    case StakingDataSource.KEY_PATH: {
         return Buffer.concat([
             uint8_to_buf(stakingChoicesEncoding[stakingChoice.type] as Uint8_t),
             path_to_buf(stakingChoice.path),
         ])
     }
-    case StakingChoiceType.BLOCKCHAIN_POINTER: {
+    case StakingDataSource.BLOCKCHAIN_POINTER: {
         return Buffer.concat([
             uint8_to_buf(stakingChoicesEncoding[stakingChoice.type] as Uint8_t),
             uint32_to_buf(stakingChoice.pointer.blockIndex),
--- a/src/interactions/serialization/addressParams.ts
+++ b/src/interactions/serialization/addressParams.ts
@@ -1,6 +1,15 @@
-import type { ParsedAddressParams, StakingChoice, Uint8_t } from "../../types/internal"
+import { ParsedAddressParams, SpendingChoice, SpendingChoiceType, StakingChoice, Uint8_t } from "../../types/internal"
 import { AddressType, StakingChoiceType } from "../../types/internal"
 import { hex_to_buf, path_to_buf, uint8_to_buf, uint32_to_buf } from "../../utils/serialize"
+
+function serializeSpendingChoice(spendingChoice: SpendingChoice): Buffer {
+    switch (spendingChoice.type) {
+    case SpendingChoiceType.PATH:
+        return path_to_buf(spendingChoice.path)
+    case SpendingChoiceType.SCRIPT_HASH:
+        return hex_to_buf(spendingChoice.scriptHash)
+    }
+}
 
 function serializeStakingChoice(stakingChoice: StakingChoice): Buffer {
     const stakingChoicesEncoding = {
@@ -49,7 +58,7 @@ export function serializeAddressParams(
         params.type === AddressType.BYRON
             ? uint32_to_buf(params.protocolMagic)
             : uint8_to_buf(params.networkId),
-        path_to_buf(params.spendingPath),
+        serializeSpendingChoice(params.spendingChoice),
         serializeStakingChoice(params.stakingChoice),
     ])
 }
--- a/src/interactions/serialization/addressParams.ts
+++ b/src/interactions/serialization/addressParams.ts
@@ -8,6 +8,8 @@ function serializeSpendingChoice(spendingChoice: SpendingChoice): Buffer {
         return path_to_buf(spendingChoice.path)
     case SpendingChoiceType.SCRIPT_HASH:
         return hex_to_buf(spendingChoice.scriptHash)
+    case SpendingChoiceType.NONE:
+        return Buffer.alloc(0)
     }
 }
 
--- a/src/interactions/serialization/addressParams.ts
+++ b/src/interactions/serialization/addressParams.ts
@@ -1,9 +1,10 @@
-import { validate } from "../../utils/parse"
-import { ParsedAddressParams, SpendingDataSource, SpendingDataSourceType, StakingDataSource, Uint8_t, Version } from "../../types/internal"
+import { InvalidDataReason } from "../../errors/invalidDataReason"
+import type { ParsedAddressParams, SpendingDataSource, StakingDataSource, Uint8_t, Version } from "../../types/internal"
+import { SpendingDataSourceType } from "../../types/internal"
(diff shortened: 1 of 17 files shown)


src/types/internal.ts
```diff
@@ -244,14 +259,31 @@ export type ShelleyAddressParams = {
         AddressType.REWARD_KEY |
         AddressType.REWARD_SCRIPT,
     networkId: Uint8_t,
-    spendingPath: ValidBIP32Path
-} & ( // Extra properties
+} & (   // TODO make "spendingChoice" object ?
```
janmazak, line 262: Already done?

src/types/internal.ts
```diff
@@ -244,14 +259,31 @@ export type ShelleyAddressParams = {
...
+            spendingChoice: SpendingChoicePath
+        } | {
+            type: AddressType.BASE_PAYMENT_SCRIPT_STAKE_KEY |
+            AddressType.BASE_PAYMENT_SCRIPT_STAKE_SCRIPT |
+            AddressType.ENTERPRISE_SCRIPT |
+            AddressType.POINTER_SCRIPT |
+            AddressType.REWARD_SCRIPT
+            spendingChoice: SpendingChoiceScriptHash
+        }
+) & ( // Extra properties
```
janmazak, line 278: Seems outdated.

src/parsing/address.ts
```diff
@@ -15,125 +126,30 @@ export function parseAddress(
...
-        const spendingPath = parseBIP32Path(params.spendingPath, InvalidDataReason.ADDRESS_INVALID_SPENDING_PATH)
-
-        return {
-            type: address.type,
-            networkId,
-            spendingPath,
-            stakingChoice: {
-                type: StakingChoiceType.NO_STAKING,
-            },
+            spendingChoice: spendingChoice as any,
```
janmazak, line 143: I don't understand why we would want to cast it to `any`. Why don't we do it as below?

janmazak:
```suggestion
            spendingChoice,
```

src/types/internal.ts
```diff
@@ -186,12 +187,31 @@ export type ParsedPoolMetadata = {
     hashHex: FixlenHexString<32>,
 } & { __brand: 'pool_metadata' }
 
+//TODO no spending for REWARD
```
janmazak, line 190: Haven't we already implemented this?

src/types/public.ts
```diff
@@ -178,9 +187,9 @@ export type AddressParamsByron = {
  * @category Addresses
  * @see [[DeviceOwnedAddress]]
  */
-export type AddressParamsBase = {
-    spendingPath: BIP32Path
-} & AddressParamsBaseStaking
+export type AddressParamsBase =
+    | { spendingPath: BIP32Path} & AddressParamsBaseStaking
+    | { spendingScriptHash: string} & AddressParamsBaseStaking
```
janmazak, line 192: Perhaps use `& AddressParamsBaseStaking` just once?

src/parsing/address.ts
```diff
@@ -1,11 +1,122 @@
...
 import { InvalidDataReason } from "../errors/invalidDataReason"
 import type { ParsedAddressParams } from "../types/internal"
-import { AddressType, KEY_HASH_LENGTH, StakingChoiceType } from "../types/internal"
+import { AddressType, KEY_HASH_LENGTH, SCRIPT_HASH_LENGTH, SpendingChoice,
+    StakingChoice, SpendingChoiceType, StakingChoiceType } from "../types/internal"
 import type { BIP32Path, BlockchainPointer, DeviceOwnedAddress, Network } from "../types/public"
 import { parseBIP32Path, parseHexStringOfLength, parseUint32_t, validate } from "../utils/parse"
 import { parseNetwork } from "./network"
 
+function extractSpendingChoice(
```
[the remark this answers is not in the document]
mkv-vcm: I'm just blending into the already existing code. Spending/StakingDataSource sounds nice.

src/parsing/address.ts
mkv-vcm: Otherwise it would not compile. By separating the creation of the SpendingChoice object from the creation of the ParsedAddressParams, the compiler cannot guarantee that the type based restrictions on the union are fulfilled.

src/types/internal.ts
mkv-vcm: Nope, just finished it an hour or two ago.

src/types/public.ts
mkv-vcm: Yeah, probably. I was still new to the TS union syntax here.

src/parsing/address.ts
```diff
@@ -1,11 +1,122 @@
...
+                type: StakingChoiceType.STAKING_KEY_PATH,
+                path: codedStakingPath,
+            }            
+        case AddressType.BASE_PAYMENT_SCRIPT_STAKE_SCRIPT:
+        case AddressType.BASE_PAYMENT_KEY_STAKE_SCRIPT:
+        case AddressType.REWARD_SCRIPT:
+            validate(stakingPath == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+            validate(stakingKeyHashHex == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+            validate(stakingBlockchainPointer == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+            validate(stakingScriptHash != null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
```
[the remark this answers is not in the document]
mkv-vcm: I really don't see how that would make the code more readable. One more layer of separation for no less code.

src/parsing/address.ts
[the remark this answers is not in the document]
janmazak: I know it has been there for a long time --- we would need to change it in lots of places (in both C and js code), but I think it is worth doing. `StakingDataSource` seems OK.

src/parsing/address.ts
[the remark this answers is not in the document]
janmazak: As it is, `extractStakingChoice` is doing two things: parsing staking choice and validating it against address type. If a function does only a single thing, it is typically easy to accurately name it (so that one does not even need to read its code or documentation when it is called from somewhere --- the name itself gives the reader all the information he needs), and usually easy to reuse it if the need arises.

Also, imagine one runs the code with the result of `InvalidDataReason.ADDRESS_INVALID_STAKING_INFO` being thrown. In the current version of the code, there are more than 10 places where that is thrown. If we could reduce it to, say, four, it would probably be easier to find the source of the exception. (In this case, one could live with it as it is because the whole thing is not that complex, but in general, having the same validation check scattered all over the place is a pattern to avoid.)

src/types/public.ts
```diff
@@ -178,9 +187,9 @@ export type AddressParamsByron = {
  * @category Addresses
  * @see [[DeviceOwnedAddress]]
  */
-export type AddressParamsBase = {
-    spendingPath: BIP32Path
-} & AddressParamsBaseStaking
+export type AddressParamsBase =
+    ({ spendingPath: BIP32Path} |
+    { spendingScriptHash: string}) & AddressParamsBaseStaking
```
janmazak, line 192:
```suggestion
export type AddressParamsBase = (
    | { spendingPath: BIP32Path}
    | { spendingScriptHash: string}
) & AddressParamsBaseStaking
```

src/parsing/address.ts
janmazak: I've consulted this with @refi93 and we don't have anything better than this, typescript is just not clever enough to deal with this properly :) Let's just add an explanatory comment on why we are doing this, and possibly you can create a very low priority issue asking for a more clever solution.

mkv-vcm: Yeah, I consulted with emo about this, and he suggested the cast to any after I explained the problem to him. And he is said to be the TS wiz in the company, so I believed him.

src/parsing/address.ts
```diff
@@ -1,139 +1,182 @@
...
         }
-    }
-    case AddressType.REWARD: {
-        validate(params.stakingBlockchainPointer == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
-        validate(params.stakingKeyHashHex == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
-        validate(params.spendingPath == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
-
-        const stakingPath = parseBIP32Path(params.stakingPath, InvalidDataReason.ADDRESS_INVALID_STAKING_KEY_PATH)
-
+    } else {
```
gabrielKerekes, line 173: In this case the nesting inside the `else` seems redundant to me.

src/types/internal.ts
```diff
@@ -200,49 +219,82 @@ type ParsedBlockchainPointer = {
...
+        AddressType.ENTERPRISE_KEY |
+        AddressType.ENTERPRISE_SCRIPT |
+        AddressType.POINTER_KEY |
+        AddressType.POINTER_SCRIPT |
+        AddressType.REWARD_KEY |
+        AddressType.REWARD_SCRIPT,
     networkId: Uint8_t,
-    spendingPath: ValidBIP32Path
-} & ( // Extra properties
+} & (
```
gabrielKerekes, line 264: IMHO it would be worth it to extract spending and staking types into helper types. E.g.
```
type ShelleyAddressParamsSpendingDataSource {
    type: AddressType.BASE_PAYMENT_KEY_STAKE_KEY |
        AddressType.BASE_PAYMENT_KEY_STAKE_SCRIPT |
        AddressType.ENTERPRISE_KEY |
        AddressType.POINTER_KEY
    spendingDataSource: SpendingDataSourcePath
} | {
    type: AddressType.BASE_PAYMENT_SCRIPT_STAKE_KEY |
    AddressType.BASE_PAYMENT_SCRIPT_STAKE_SCRIPT |
    AddressType.ENTERPRISE_SCRIPT |
    AddressType.POINTER_SCRIPT
    spendingDataSource: SpendingDataSourceScriptHash
} | {
    type: AddressType.REWARD_KEY | AddressType.REWARD_SCRIPT
    spendingDataSource: SpendingDataSourceNone
}
```
But I would name it `ShelleyAddressParamsSpendingInfo`.

src/types/public.ts
```diff
@@ -18,30 +18,36 @@ export type bigint_like = number | bigint | string
  */
 export enum AddressType {
     /**
-     * Shelley base address
+     * Shelley base addresses
      * @see [[AddressParamsBase]]
      */
-    BASE = 0b0000,
+    BASE_PAYMENT_KEY_STAKE_KEY          = 0b0000,
```
gabrielKerekes, line 24: I would suggest keeping the old names of the addresses the same and just adding `SCRIPT` suffix to the script address types. In the case of Base address I would use:
```
BASE,
BASE_SCRIPT_KEY,
BASE_KEY_SCRIPT,
BASE_SCRIPT_SCRIPT
```

gabrielKerekes: Also, I don't the aligning to the right is necessary.

src/types/public.ts
```diff
@@ -178,9 +187,10 @@ export type AddressParamsByron = {
  * @category Addresses
  * @see [[DeviceOwnedAddress]]
  */
-export type AddressParamsBase = {
-    spendingPath: BIP32Path
-} & AddressParamsBaseStaking
+export type AddressParamsBase = (
+    { spendingPath: BIP32Path} |
+    { spendingScriptHash: string}
+) & AddressParamsBaseStaking
```
gabrielKerekes, line 193: Since we have `AddressParamsBaseStaking` I'd introduce `AddressParamsBaseSpending` now.

src/types/public.ts
```diff
@@ -190,26 +200,31 @@ export type AddressParamsBase = {
...
 /**
  * Shelley *pointer* address parameters
  * @category Addresses
  * @see [[DeviceOwnedAddress]]
  * */
-export type AddressParamsPointer = {
+// TODO finish the transformations
+export type AddressParamsPointer = ({
     spendingPath: BIP32Path
+} | {
```
gabrielKerekes, line 225: I'd extract this to `AddressParamsPointerSpending`.

src/utils.ts
```diff
@@ -105,7 +105,8 @@ function getShelleyAddressPrefix(data: Buffer): string {
 
     const addressType = (data[0] & 0b11110000) >> 4
     switch (addressType) {
-    case AddressType.REWARD:
+    case AddressType.REWARD_KEY:
```
gabrielKerekes, line 108: Do we need to have these functions duplicated here if they are exactly the same in `src/utils/address.ts`? @janmazak maybe?

src/interactions/serialization/addressParams.ts
```diff
@@ -1,34 +1,50 @@
...
+import { AddressType, StakingDataSourceType } from "../../types/internal"
 import { hex_to_buf, path_to_buf, uint8_to_buf, uint32_to_buf } from "../../utils/serialize"
+import { getCompatibility } from "../getVersion"
+import { InvalidDataReason } from "../../errors/invalidDataReason"
 
-function serializeStakingChoice(stakingChoice: StakingChoice): Buffer {
+function serializeSpendingChoice(spendingChoice: SpendingDataSource): Buffer {
+    switch (spendingChoice.type) {
+    case SpendingDataSourceType.PATH:
+        return path_to_buf(spendingChoice.path)
```
gabrielKerekes, line 11: Wouldn't it be better to also encode the `spendingChoice.type` as is done with `stakingChoice`? IMHO it would make the code more consistent also on Ledger app side - both `spendingChoice` and `stakingChoice` would be handled similarly then.

src/parsing/address.ts
```diff
@@ -1,139 +1,182 @@
...
-    const params = address.params as {
-        spendingPath?: BIP32Path,
-        stakingPath?: BIP32Path
-        stakingKeyHashHex?: string
-        stakingBlockchainPointer?: BlockchainPointer
+function extractSpendingDataSource(
+    spendingPath?: BIP32Path,
+    spendingScriptHash?: string,
+): SpendingDataSource {
+    if (null != spendingPath) {
```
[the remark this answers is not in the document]
mkv-vcm: I would be curious to get a set of different opinions for this one, but I guess a "coding guideline" would be way too formal for a project like this.

src/types/public.ts
mkv-vcm: Why would you want that? I don't understand why less information would be good here and why BASE should be special.

Aligning is good if you want to do block operations.

src/types/public.ts
mkv-vcm: Rather, we should eliminate the AddressParamsBaseStaking.

src/interactions/serialization/addressParams.ts
mkv-vcm: The spending source is clear from the type of the address. The staking is not. So for spending it would be purely redundant information, which just leads to more chance to have contradictory information.

src/types/public.ts
gabrielKerekes: By keeping the old names we would have some backwards compatibility so users wouldn't have to change their current code. Base address is special because it has multiple variations and we need to distinguish them. Also the shorter BASE names seem less confusing to me - but that's maybe just because the names are already used on Trezor.

What do you mean by block operations? Since the aligning isn't used elsewhere I would also drop it here, IMHO it can only cause issues when renaming the code or something like that.

Anyways, I think we should keep the original names of the older address types. The rest are just suggestions :wink:

src/types/public.ts
gabrielKerekes: I'm not a fan of the nested types. IMHO
```
export type AddressParamsBase =
& AddressParamsBasePayment
& AddressParamsBaseStaking
```
reads better than
```
export type AddressParamsBase = (
    { spendingPath: BIP32Path} |
    { spendingScriptHash: string}
) & (
    { stakingPath: BIP32Path } |
    { stakingKeyHashHex: string } |
    { stakingScriptHash: string} |
)
```
But I guess it's subjective so we can leave it.

mkv-vcm:
Included in the multisig PR and all concerns are resolved or addressed there.

PR #116, opened 2021-06-25
Add native script derivation

KubqoA:
# Native script derivation

*Keeping in Draft state as the Ledger Cardano app doesn't yet support this feature*

Given a script in the form defined as [NativeScript](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/add-script-derivation/src/types/public.ts#L1133) returns a script hash in form of a hex encoded string.

The `INS` assigned to the derive script hash call is `0x21`.

After parsing the `NativeScript` I split the resulting `ParsedNativeScript` into two subsets.
1. `ParsedComplexNativeScript` are scripts that have nested "subscripts" (types: `ALL`, `ANY`, `N_OF_K`).
    The interaction on ledger should display the series of messages:
    * Information about the script (e.g. *Script - ALL, Confirm 12 scripts*)
    * Then each individual script
    * And a final *finished* message
   
    The calls to Ledger reflect this structure:
    * The call is initiated with `P1=0x01` and sends the type of script in a number (*uint8*) representation according to [the cddl](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L242)
     , the number (*uint32*) of all subscripts contained within the script, and in case of script type `N_OF_K` also the number (*uint32*) of required scripts.
    * Then the individual subscripts are processed

    It's Ledger's responsibility to keep track of the current depth and index of currently processed script.

2. `ParsedSimpleNativeScript` are scripts without "subscripts" (types: `PUBKEY_DEVICE_OWNED`, `PUBKEY_THIRD_PARTY`, `INVALID_BEFORE`, `INVALID_HEREAFTER`).
    The interaction on ledger should only display one message containing the information about the script.
    For example: *Script - Key hash, 80f9e2c88e6c817008f3a812ed889b4a4da8e0bd103f86e7335422aa*

    Therefore for this subset of scripts only call to ledger is made:
    The call initiated with `P1=0x02`. The data sent begins with a number (*uint8*) representation of the script type
    and then one of the following depending on the type:
    * For `PUBKEY_DEVICE_OWNED` first a byte (*uint8*) of value `1`, to distinguish between the two `PUBKEY` script types, and then it's the byte representation of the public key path
    * For `PUBKEY_THIRD_PARTY` first a byte (*uint8*) of value `2`, to distinguish between the two `PUBKEY` script types, and then it's the byte representation of the provided key hash
    * For `INVALID_BEFORE` and `INVALID_HEREAFTER` a number (*uint64*)

After the whole native script is sent to the Ledger device it will send an explicit finish call (`P1 = 0x03`) with one *uint8* value specifying the display format of the native script hash on the Ledger.


src/interactions/common/ins.ts
```diff
@@ -4,6 +4,7 @@ export const enum INS {
 
   GET_EXT_PUBLIC_KEY = 0x10,
   DERIVE_ADDRESS = 0x11,
+  DERIVE_SCRIPT_HASH = 0x12,
```
KubqoA, line 7: I wasn't sure, if there was some specification or pattern by which we assign the `INS`.

src/interactions/getVersion.ts
```diff
@@ -40,6 +40,8 @@ export function getCompatibility(version: Version): DeviceCompatibility {
     const v2_2 = isLedgerAppVersionAtLeast(version, 2, 2) && isLedgerAppVersionAtMost(version, 2, Infinity)
     const v2_3 = isLedgerAppVersionAtLeast(version, 2, 3) && isLedgerAppVersionAtMost(version, 2, Infinity)
     const v2_4 = isLedgerAppVersionAtLeast(version, 2, 4) && isLedgerAppVersionAtMost(version, 2, Infinity)
+    // TODO: specify the exact version of ledger-app to support script derivation
+    const v2_x = isLedgerAppVersionAtLeast(version, 2, 5) && isLedgerAppVersionAtMost(version, 2, Infinity)
```
KubqoA, line 44: The exact version should probably be specified once we know what release of the Ledger app will support this functionality.

src/parsing/script.ts
```diff
@@ -0,0 +1,76 @@
...
+        return {
+            type,
+            params: {
+                keyHashHex,
+            },
+        }
+    }
+    case ScriptType.ALL:
+    case ScriptType.ANY: {
+        validate(isArray(params.scripts), InvalidDataReason.DERIVE_SCRIPT_HASH_SCRIPTS_NOT_AN_ARRAY)
```
gabrielKerekes, line 34: I think we should also validate that the array is not empty as it IMHO doesn't make sense to have an empty script. The same for `N_OF_K`.

src/types/internal.ts
```diff
@@ -266,3 +266,37 @@ export type ParsedOperationalCertificate = {
     issueCounter: Uint64_str,
     coldKeyPath: ValidBIP32Path,
 }
+
+export const SCRIPT_HASH_LENGTH = 28
+
+export type ParsedSimpleScript = {
+    type: ScriptType.PUBKEY,
+    params: {
+        keyHashHex: FixlenHexString<typeof KEY_HASH_LENGTH>,
```
gabrielKerekes, line 275: The key also be sent as a path if it's known to be owned by the user/device. Similarly to `PoolKey` - https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/develop/src/types/public.ts#L368.

src/interactions/deriveScriptHash.ts
```diff
@@ -0,0 +1,112 @@
...
+        expectedResponseLength: expectsResponse ? SCRIPT_HASH_LENGTH : 0,
+    })
+    return responseIf(expectsResponse, response)
+}
+
+function hasScriptSubscripts(script: ParsedScript): script is ParsedComplexScript {
+    return script.type === ScriptType.ALL || script.type === ScriptType.ANY || script.type === ScriptType.N_OF_K
+}
+
+function *deriveScriptHash_rec<B extends boolean>(
```
gabrielKerekes, line 77: I would rather call this `deriveScriptHash_addScript`. And the other functions similarly:
`deriveScriptHash_scriptHeader` -> `deriveScriptHash_addScriptHeader`
`deriveScriptHash_scriptFinished` -> `deriveScriptHash_finishScript`
`deriveScriptHash_script` -> `deriveScriptHash_addSimpleScript`

src/interactions/deriveScriptHash.ts
```diff
@@ -0,0 +1,112 @@
...
+    const response = yield send({
+        p1: P1.STAGE_SCRIPT_FINISHED,
+        p2: P2.UNUSED,
+        data: serializeScriptFinished(script),
+        expectedResponseLength: expectsResponse ? SCRIPT_HASH_LENGTH : 0,
+    })
+    return responseIf(expectsResponse, response)
+}
+
+function hasScriptSubscripts(script: ParsedScript): script is ParsedComplexScript {
```
gabrielKerekes, line 73: This can IMHO be easily called `hasSubscripts`. It's private to this file and the first parameter is `script` so I think it's OK to leave `Script` out from the name.

src/interactions/deriveScriptHash.ts
```diff
@@ -0,0 +1,112 @@
...
+    STAGE_SCRIPT_HEADER = 0x01,
+    STAGE_SCRIPT = 0x02,
+    STAGE_SCRIPT_FINISHED = 0x03,
+}
+
+const enum P2 {
+    UNUSED = 0x00,
+}
+
+type ScriptHashResponse<B> = B extends true ? { scriptHashHex: string } : void
```
gabrielKerekes, line 28: I don't enjoy this 😄  But I don't have a better solution yet.

gabrielKerekes (commented):
I'm not sure what stage this is in, but it looks good so far 😉  I feel like this could already work as it is, no?

src/parsing/script.ts
```diff
@@ -0,0 +1,76 @@
...
+        invalidBefore?: bigint_like,
+        invalidHereafter?: bigint_like,
+        scripts?: NativeScript[],
+    }
+
+    const type = script.type
+
+    switch (type) {
+    case ScriptType.PUBKEY: {
+        const keyHashHex = parseHexStringOfLength(params.keyHash, KEY_HASH_LENGTH, InvalidDataReason.DERIVE_SCRIPT_HASH_INVALID_KEY_HASH)
```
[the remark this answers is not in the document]
KubqoA: Yeah I wasn't sure, whether this was necessary, but I'll add it.

src/interactions/deriveScriptHash.ts
KubqoA: Yeah it's basically a workaround, to have a correct TypeScript type, and less duplication of code, for the `deriveScriptHash_*` calls.

src/parsing/script.ts
KubqoA: Do you think we should also validate for `N_OF_K` that the `N` is less or equal than `K`? e.g. we don't permit requiring more scripts than there are available?

src/types/internal.ts
KubqoA: Okay, I see it also in the docs, probably overlooked it the first time, will add it later :)

src/parsing/script.ts
gabrielKerekes: Good point, but I think this is more of a job for Ledger than for LedgerJS. IMHO LedgerJS should validate that the data have the right structure i.e. that it is sendable to Ledger, but not that it is "semantically" correct.

src/interactions/serialization/script.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export function serializeScript(
+    script: ParsedSimpleScript
+): Buffer {
+    switch (script.type) {
+    case ScriptType.PUBKEY_DEVICE_OWNED:
+        return Buffer.concat([
+            uint8_to_buf(TYPE_ENCODING[script.type]),
+            uint8_to_buf(0 as Uint8_t), // Byte to distinguish path from hash
```
KubqoA, line 44: @gabrielKerekes Here I am not 100% sure this is the correct way to do this. I tried to keep the `TYPE_ENCODING` in accord with the cddl, but the cddl only specifies the `pubkey` script as having an `addr_keyhash` ([line 241](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L241)).
So my guess is it should also be okay to modify the `TYPE_ENCODING` to something like this, as this is an information primarily intended for the ledger app to distinguish what type of script it is processing:
```ts
const TYPE_ENCODING = {
    [ScriptType.PUBKEY_DEVICE_OWNED]: 0 as Uint8_t,
    [ScriptType.PUBKEY_THIRD_PARTY]: 1 as Uint8_t,
    [ScriptType.ALL]: 2 as Uint8_t,
    [ScriptType.ANY]: 3 as Uint8_t,
    [ScriptType.N_OF_K]: 4 as Uint8_t,
    [ScriptType.INVALID_BEFORE]: 5 as Uint8_t,
    [ScriptType.INVALID_HEREAFTER]: 6 as Uint8_t,
}
```
and drop this one extra byte, but again I wasn't sure it was the right thing to do, so I would like to hear somebody else's opinion.

src/interactions/deriveScriptHash.ts
KubqoA: I tried to think about the naming and in the end went with:
`deriveScriptHash_scriptHeader` -> `deriveScriptHash_startComplexScript`
`deriveScriptHash_scriptFinished` -> `deriveScriptHash_finishComplexScript`
`deriveScriptHash_script` -> `deriveScriptHash_addSimpleScript`

gabrielKerekes: 👌

src/interactions/serialization/script.ts
gabrielKerekes: Even though only the `PUBKEY` script contains a key I think it makes sense to handle this similarly to `addressParams`. So there would only be a `ScriptType.PUBKEY`. The public params would be of type `ScriptParamsDeviceOwnedPubkey | ScriptParamsThirdPartyPubkey` or simply as is done with `AddressParamsBaseStaking`. This would get parsed into `ParsedSimpleScript.PUBKEY` which would additionally contain a `ScriptKeyType` which could either be `THIRD_PARTY` or `DEVICE_OWNED` (similarly to `ParsedPoolOwner` or other types). This would then be serialized into Ledger instead of the `0` and `1` you are serializing now. I am not sure if it's OK to serialize it just for the PUBKEY script type or whether it would need to be included for all scripts even though they don't need it.

We should probably also get @janmazak 's opinion on this.

src/parsing/script.ts
mkv-vcm: I think both of these are semantic questions. Empty arrays don't make sense, but they are not structurally impossible.

src/types/internal.ts
```diff
@@ -266,3 +266,42 @@ export type ParsedOperationalCertificate = {
     issueCounter: Uint64_str,
     coldKeyPath: ValidBIP32Path,
 }
+
+export const SCRIPT_HASH_LENGTH = 28
```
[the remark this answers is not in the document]
KubqoA: Thanks for the heads up.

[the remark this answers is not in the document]
KubqoA: In the last commit I renamed it to `NATIVE_SCRIPT_HASH_LENGTH`.

src/interactions/serialization/nativeScript.ts
```diff
@@ -0,0 +1,72 @@
...
+    case NativeScriptType.ANY:
+        return Buffer.concat([
+            uint8_to_buf(TYPE_ENCODING[script.type]),
+            uint32_to_buf(script.params.scripts.length as Uint32_t),
+        ])
+    case NativeScriptType.N_OF_K:
+        return Buffer.concat([
+            uint8_to_buf(TYPE_ENCODING[script.type]),
+            uint64_to_buf(script.params.requiredCount), // N
+            uint32_to_buf(script.params.scripts.length as Uint32_t), // K
```
KubqoA, line 30: @gabrielKerekes @mkv-vcm Here I am not sure, whether these are correct sizes of the `uint`. In the cddl the `n_of_k` script is specified as follows:
```
script_n_of_k = (3, n: uint, [ * native_script ])
```
The `uint` there if I am not mistaken explicitly means a `uint64`.

However, when I am serializing the length/size of the `native_script` array, it's getting serialized as `uint32`, because when I looked at [other parts of the codebase](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/develop/src/interactions/serialization/txInit.ts#L41) I saw that that's the way to serialize the length of an array.

But it doesn't really make sense to serialize the `N` parameter as a `uint64` and the size (`K`) as `uint32`, because semantically, the `N` should be less or equal to `K`. What do you think is the correct approach here?

gabrielKerekes: I would say let's have both as `uint32`. IHMO it should be enough to have 4294967295 possible required/total signatures 😄 It shouldn't cause any practical problems to users.

[Transaction input index](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/develop/src/interactions/serialization/txOther.ts#L9) is also a `uint32` even though the [CDDL mentions](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L63) `uint`.

KubqoA: Okay, fixed it as you suggested. Thanks!

src/interactions/common/ins.ts
janmazak: Not really.

src/parsing/script.ts
KubqoA: @gabrielKerekes actually just checked with `cardano-cli` and it accepts for example an `ALL` script with no nested scripts and generates a hash for it.

src/interactions/getVersion.ts
KubqoA: @janmazak Is it okay to suppose the version of Ledger app which will include the native script hash derivation to be at least 2.5?
E.g. can I change it here from `v2_x` to `v2_5`?

src/interactions/deriveScriptHash.ts
KubqoA: Solved when I added the finishWholeNativeScript call :)

src/parsing/script.ts
KubqoA: I suggest we remove the check for empty arrays, as `cardano-cli` allows it. It doesn't really make any sense to derive a script hash for such a script, but it's possible.

janmazak: Yes, we should support  empty scripts; confirmed by Duncan.

src/interactions/deriveNativeScriptHash.ts
```diff
@@ -0,0 +1,119 @@
...
+    yield send({
+        p1: P1.STAGE_COMPLEX_SCRIPT_FINISH,
+        p2: P2.UNUSED,
+        data: serializeComplexNativeScriptFinish(script),
+        expectedResponseLength: 0,
+    })
+}
+
+function isComplexScript(script: ParsedNativeScript): script is ParsedComplexNativeScript {
+    return script.type === NativeScriptType.ALL || script.type === NativeScriptType.ANY || script.type === NativeScriptType.N_OF_K
```
janmazak, line 65: I think a switch statement would be easier to comprehend.

src/parsing/nativeScript.ts
```diff
@@ -0,0 +1,129 @@
...
+    const params = script.params as {
+        path?: BIP32Path,
+        keyHashHex?: string,
+        requiredCount?: bigint_like,
+        invalidBefore?: bigint_like,
+        invalidHereafter?: bigint_like,
+        scripts?: NativeScript[],
+    }
+
+    const type = script.type
```
janmazak, line 21: I'm not sure this makes it easier to read (because "type" is not specific enough, and `script.type` is not unreasonably long).

src/parsing/nativeScript.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    switch (type) {
+    case NativeScriptType.PUBKEY_DEVICE_OWNED: {
+        validate(params.keyHashHex == null, InvalidDataReason.DERIVE_NATIVE_SCRIPT_HASH_INVALID_DATA)
+        validate(params.requiredCount == null, InvalidDataReason.DERIVE_NATIVE_SCRIPT_HASH_INVALID_DATA)
+        validate(params.invalidBefore == null, InvalidDataReason.DERIVE_NATIVE_SCRIPT_HASH_INVALID_DATA)
+        validate(params.invalidHereafter == null, InvalidDataReason.DERIVE_NATIVE_SCRIPT_HASH_INVALID_DATA)
+        validate(params.scripts == null, InvalidDataReason.DERIVE_NATIVE_SCRIPT_HASH_INVALID_DATA)
```
janmazak, line 29: I find this redundancy annoying. The error and the check are always the same, why bother reading it line by line?... What about having a function that checks that every field of an object is null except one? (Or except a given list, since one time we need two?) Perhaps consult with @refi93 (or some other typescript expert via dev-queries) whether such functions are a good practice or not. Such a function will probably be useful elsewhere in this repository.

test/integration/__fixtures__/deriveNativeScriptHash.ts
```diff
@@ -0,0 +1,390 @@
...
+                            keyHashHex: '3g55d9f68255dfbefa1efd711f82d005fae1be2e145d616c90cf0fa9',
+                        },
+                    },
+                ],
+            },
+        },
+        invalidDataReason: InvalidDataReason.DERIVE_NATIVE_SCRIPT_HASH_INVALID_REQUIRED_COUNT,
+    },
+    {
+        testname: "N_OF_K - invalid required count (negative number)",
```
janmazak, line 350: Isn't this the same as the previous test? Just with a different keyHashHex.

test/integration/deriveNativeScriptHash.test.ts
```diff
@@ -0,0 +1,59 @@
...
+                    script,
+                    displayFormat,
+                })
+
+                expect(scriptHashHex).to.equal(expectedHash)
+            })
+        }
+    })
+
+    describe("Invalid scripts on Ledger", async () => {
```
janmazak, line 35: Why is "on Ledger" included in the name? Are these tests supposed to pass the typescript validation, but fail on Ledger? Typically, that can only happen because of security policies that are not enforced by ledgerjs (the js validation should be "complete").

test/integration/__fixtures__/deriveNativeScriptHash.ts
KubqoA: Yes I was just cleaning up the codebase and noticed some tests were duplicated.

src/parsing/nativeScript.ts
KubqoA: I did this based on @gabrielKerekes [suggestion](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/116#discussion_r662795610)

test/integration/deriveNativeScriptHash.test.ts
KubqoA: I based this on the fact that @gabrielKerekes [suggested that some validation can be left for Ledger](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/116#discussion_r662942876).

src/errors/invalidDataReason.ts
```diff
@@ -156,4 +156,14 @@ export enum InvalidDataReason {
...
   OPERATIONAL_CERTIFICATE_INVALID_COLD_KEY_PATH =
   "invalid operational certificate cold key path",
+
+  DERIVE_NATIVE_SCRIPT_HASH_INVALID_DATA = "invalid native script input",
+  DERIVE_NATIVE_SCRIPT_HASH_INVALID_KEY_PATH = "invalid key path param",
+  DERIVE_NATIVE_SCRIPT_HASH_INVALID_KEY_HASH = "invalid key hash param",
+  DERIVE_NATIVE_SCRIPT_HASH_SCRIPTS_NOT_AN_ARRAY = "invalid scripts - scripts is not an array",
+  DERIVE_NATIVE_SCRIPT_HASH_INVALID_REQUIRED_COUNT = "invalid required count",
+  DERIVE_NATIVE_SCRIPT_HASH_INVALID_INVALID_BEFORE = "invalid invalid before",
+  DERIVE_NATIVE_SCRIPT_HASH_INVALID_INVALID_HEREAFTER = "invalid invalid hereafter",
```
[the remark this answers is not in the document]
KubqoA: One solution I can think of, is to name the parameter `slot` (as it is named when using the `cardano-cli`) and then merge these two invalid reasons to one named `DERIVE_NATIVE_SCRIPT_HASH_INVALID_SLOT`. Do you think that would be better?

src/parsing/nativeScript.ts
KubqoA: Tried to create a solution for this, see if you like it better :)

src/interactions/deriveNativeScriptHash.ts
KubqoA: I implemented it via a switch statement, imho the previous version was better, but see for yourself. Maybe you had something different in mind?

test/integration/deriveNativeScriptHash.test.ts
KubqoA: But I see that this should be catched by ledgerjs as well. Will implement

src/interactions/deriveNativeScriptHash.ts
janmazak: I definitely like the switch better, no need to worry about paranthesizing complex expressions or pondering about == vs. === :)

src/utils.ts
```diff
@@ -125,6 +125,17 @@ export function bech32_decodeAddress(data: string): Buffer {
     return Buffer.from(bech32.fromWords(words))
 }
 
+export function array_symmetricDifference(arr1: any[], arr2: any[]): any[] {
+    let diffrence = new Set(arr1)
```
janmazak, line 129:
```suggestion
    let difference = new Set(arr1)
```
Or just `diff`.

src/utils.ts
```diff
@@ -125,6 +125,17 @@ export function bech32_decodeAddress(data: string): Buffer {
     return Buffer.from(bech32.fromWords(words))
 }
 
+export function array_symmetricDifference(arr1: any[], arr2: any[]): any[] {
```
janmazak, line 128: I'm not sure about conventions on naming functions in this repo.

src/parsing/nativeScript.ts
janmazak: I like it much better. Does someone really familiar with js/ts approved this? I am never sure objects contain only the keys I think they contain, and not a load of stuff inherited from some prototype or what.

It will probably come in handy in other places, e.g. `parseAddress`; you might want to discuss this with Marton.

src/errors/invalidDataReason.ts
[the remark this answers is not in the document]
KubqoA: The `invalid_before` and `invalid_hereafter` would be both named `slot`, that's also the way the `cardano-cli` named it

src/parsing/nativeScript.ts
KubqoA: No I haven't consulted it with anyone else, should I consult it with Rafo?

janmazak: Definitely, we might learn something and the code won't be worse off if we ask :)

src/errors/invalidDataReason.ts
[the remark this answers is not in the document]
janmazak: Yes, let's use `slot`.

src/utils.ts
```diff
@@ -125,6 +125,17 @@ export function bech32_decodeAddress(data: string): Buffer {
...
+    let diff = new Set(arr1)
+    new Set(arr2).forEach(item => diff.has(item) ?
+        diff.delete(item) : diff.add(item))
+    return [...diff]
+}
+
+export function objectHasOnlyTheseKeys(object: object, keys: string[]): boolean {
+    return array_symmetricDifference(Object.keys(object), keys).length === 0
```
refi93, line 136: I think such strict check can become too restrictive for integrating parties, a similar problem emerged in trezor-connect where their validation helpers are enforcing presence of a strict set of fields: https://github.com/trezor/connect/issues/770

But for the integrating parties it may be convenient to set some fields to null/empty list/empty map even if they are not strictly required for the given type, especially if they have some code that tries to be agnostic to the exact data type being dealt with at runtime - this may or may not be a good design decision on the side of the integrating but our library probably shouldn't dictate that.

So checking for a whitelist of properties is imho not a good idea and can lead to unexpected issues on the end of the integrating parties which would moreover run into these checks only at runtime, so it can became pretty painful if they don't have the logic fully covered by (unit)tests (which we again shouldn't force them to) and may be forced to implement quite awkward workarounds like "manually" deleting null/empty keys before passing the object to ledgerjs - something we had to do e.g. in the case of trezor-connect

Checking for a blacklist of properties and that they are not set to anything "non-empty" (whatever that means for the specific type - may be null/empty list/map) seems more sensible and I don't see a sharply better approach than just checking ad-hoc for these keys wherever we think the caller may otherwise shoot themselves in the foot

src/parsing/nativeScript.ts
KubqoA: Based on Rafo's comment: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/116#discussion_r689531708 I decided to revert the changes.

janmazak:
WIll be merged via #130.

PR #119, opened 2021-07-05
Certificates

mkv-vcm:
Certificates multisigging JS side.

### DIFF
--- a/src/interactions/serialization/txCertificate.ts
+++ b/src/interactions/serialization/txCertificate.ts
@@ -1,24 +1,78 @@
-import type { ParsedCertificate, Uint8_t } from "../../types/internal"
+import { InvalidDataReason } from "../../errors/invalidDataReason"
+import { validate } from "../../utils/parse"
+import { CertificateIdentifier2, CertificateIdentifierType, ParsedCertificate, Uint8_t } from "../../types/internal"
 import { CertificateType } from "../../types/internal"
 import { unreachable } from "../../utils/assert"
 import { hex_to_buf, path_to_buf, uint8_to_buf, uint64_to_buf } from "../../utils/serialize"
 
+export function serializeTxCertificatePreMultisig(
+    certificate: ParsedCertificate,
+) {
+    switch (certificate.type) {
+    case CertificateType.STAKE_REGISTRATION:
+    case CertificateType.STAKE_DEREGISTRATION: {
+        validate(CertificateIdentifierType.KEY_PATH == certificate.identifier.type, InvalidDataReason.CERTIFICATE_INVALID_IDENTIFIER)
+        return Buffer.concat([
+            uint8_to_buf(certificate.type as Uint8_t),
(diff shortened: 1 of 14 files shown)


src/types/public.ts
```diff
@@ -665,14 +666,21 @@ export type PoolRetirementParams = {
     retirementEpoch: bigint_like,
 }
 
+export type CertificateIdentifier = {
```
gabrielKerekes, line 669: I think we could call this as it is in the CDDL:`StakeCredential`?

src/types/internal.ts
```diff
@@ -30,15 +30,29 @@ export const VRF_KEY_HASH_LENGTH = 32
 export const REWARD_ACCOUNT_HEX_LENGTH = 29
 export const ED25519_SIGNATURE_LENGTH = 64
 
+export const enum CertificateIdentifierType {
+    KEY_PATH = 0,
+    SCRIPT_HASH = 1,
+}
+
+export type CertificateIdentifier2 =
```
gabrielKerekes, line 38: `ParsedStakeCredential`?

src/interactions/signTx.ts
```diff
@@ -645,6 +662,11 @@ export function* signTransaction(version: Version, request: ParsedSigningRequest
         witnesses.push(witness)
     }
 
+    for (const multisigPath of multisigWitnessPaths) {
```
gabrielKerekes, line 665: Hm, I'm not sure how exactly the `multisigWitnessPaths` should be handled. It would be nicest if they could simply be part of `witnessPaths`, but that way I think we wouldn't be able to correctly validate the witnesses. We will probably require a separate call/STAGE so that the Ledger app can correctly validate both the ordinary and multisig witnesses.

src/interactions/serialization/txCertificate.ts
```diff
@@ -1,24 +1,78 @@
-import type { ParsedCertificate, Uint8_t } from "../../types/internal"
+import { InvalidDataReason } from "../../errors/invalidDataReason"
+import { validate } from "../../utils/parse"
+import { CertificateIdentifier2, CertificateIdentifierType, ParsedCertificate, Uint8_t } from "../../types/internal"
 import { CertificateType } from "../../types/internal"
 import { unreachable } from "../../utils/assert"
 import { hex_to_buf, path_to_buf, uint8_to_buf, uint64_to_buf } from "../../utils/serialize"
 
-export function serializeTxCertificate(
+export function serializeTxCertificatePreMultisig(
```
gabrielKerekes, line 8: This should probably be used in `signTx_addCertificate`.

src/interactions/signTx.ts
mkv-vcm: Who can make this decision?

src/interactions/serialization/txCertificate.ts
mkv-vcm: Added this handling in the withdrawal commit.

PR #120, opened 2021-07-28
multisig JM WIP

janmazak:


src/utils/serialize.ts
```diff
@@ -46,7 +43,17 @@ export function buf_to_uint32(data: Buffer): Uint32_t {
 export function uint64_to_buf(value: Uint64_str): Buffer {
     assert(isUint64str(value), 'invalid uint64_str')
 
-    const data = bs10.decode(value)
+    const data = new Uint64BE(value, 10).toBuffer()
+    assert(data.length <= 8, "excessive data")
+
+    const padding = Buffer.alloc(8 - data.length)
+    return Buffer.concat([padding, data])
```
janmazak, line 50: Do we need this? I don't see how an uint64 value can correctly serialize to a buffer longer than 8 bytes, and I would expect to always get 8 bytes, not a shorter buffer, as the default behavior.

src/types/public.ts
```diff
@@ -995,7 +1022,17 @@ export type Transaction = {
...
+    validityIntervalStart?: bigint_like | null,
+    /**
+     * Mint or burn instructions (if any).
+     * Assets to be minted (token amount positive) or burned (token amount negative) with the transaction.
+     * If not null, the entries' keys (policyIds) must be sorted lowest value to highest to reflect a canonical CBOR.
+     * The sorting rules (as described in the [CBOR RFC](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9)) are:
+     *  * if two keys have different lengths, the shorter one sorts earlier;
+     *  * if two keys have the same length, the one with the lower value in lexical order sorts earlier.
+     */
+    // TODO update the docstring when CIP-0021 is approved
```
janmazak, line 1034: This comment assumes enforcing canonical form which is not applied in the code currently.
We did not yet receive IOHK approval on that (see the slack discussion).

src/types/public.ts
```diff
@@ -190,26 +204,25 @@ export type AddressParamsBase = {
  */
 // Not really worth the effort of disambiguation through additional tagged enum
```
janmazak, line 205: I am not sure this is still true. I find the address params types messy. The staking part that goes into base address is the same as what goes into a reward address (maybe one of the options is forbidden in one of those places), but the types completely disregard that. We should get the opinion of Adalite's people who are integrating this stuff.

janmazak:
Will be merged via #121 .

PR #121, opened 2021-08-11
multisig

mkv-vcm:


src/interactions/signTx.ts
```diff
@@ -593,16 +668,29 @@ export function* signTransaction(version: Version, request: ParsedSigningRequest
...
+    if (scriptWitnessPaths.length == 0) {
+        for (const path of witnessPaths) {
+            const witness = yield* signTx_getWitness(path)
+            witnesses.push(witness)
+        }
+    } else {
+        for (const path of scriptWitnessPaths) {
+            const witness = yield* signTx_getWitness(path)
+            witnesses.push(witness)
+        }
```
janmazak, line 690: This seems to be not entirely consistent with what Trezor does, see `_get_witness_requests`  in 
https://github.com/vacuumlabs/trezor-firmware/pull/66/files#diff-aa355fcce844deffafbd696c1d414eb1c0198b0af2726710c7b8d8f7fda577cdR547-R582
For instance, a minting witness for an ordinary transaction will block processing of input witnesses inferred from tx body, won't it?

Let's discuss this in detail on a standup again to make sure we the same API and interpretation for both hw wallets and we have documented it properly.

src/parsing/address.ts
```diff
@@ -1,11 +1,147 @@
...
+function extractStakingDataSource(
+    stakingPath?: BIP32Path,
+    stakingKeyHashHex?: string,
+    stakingBlockchainPointer?: BlockchainPointer,
+    stakingScriptHash?: string,
+) : StakingDataSource {
+    if (null != stakingPath) {
+        validate(stakingKeyHashHex == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+        validate(stakingBlockchainPointer == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+        validate(stakingScriptHash == null, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
```
janmazak, line 42: @KubqoA has a function on his branch that deals with validating that all other fields are null, I think we want to use it here, too.

src/interactions/signTx.ts
```diff
@@ -42,6 +43,7 @@ function* signTx_init(
     tx: ParsedTransaction,
     signingMode: TransactionSigningMode,
     wittnessPaths: ValidBIP32Path[],
```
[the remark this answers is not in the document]
mkv-vcm: Oops.

src/interactions/serialization/txCertificate.ts
```diff
@@ -1,24 +1,62 @@
...
 
-export function serializeTxCertificate(
+export function serializeTxCertificatePreMultisig(
     certificate: ParsedCertificate,
 ) {
-
     switch (certificate.type) {
     case CertificateType.STAKE_REGISTRATION:
     case CertificateType.STAKE_DEREGISTRATION: {
+        validate(StakeCredentialType.KEY_PATH == certificate.stakeCredential.type, InvalidDataReason.CERTIFICATE_INVALID_STAKE_CREDENTIAL)
```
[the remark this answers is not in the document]
mkv-vcm: Done, in all of the ==s from develop till HEAD.

src/interactions/serialization/addressParams.ts
```diff
@@ -1,53 +1,92 @@
...
+            validate(StakingDataSourceType.KEY_PATH == staking.type, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
+            spending = {
+                type: SpendingDataSourceType.PATH,
+                path: staking.path,
+            }
+            staking = {
+                type: StakingDataSourceType.NONE,
+            }
+        } else if (AddressType.REWARD_SCRIPT == params.type) {
+            // TODO we do not support this (sending hashes in the spending part) on ledger at all
```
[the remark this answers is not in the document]
mkv-vcm: Not sure what the exact error should be thrown, but there it is.

src/interactions/signTx.ts
```diff
@@ -42,6 +43,7 @@ function* signTx_init(
...
+    scriptWitnessPaths: ValidBIP32Path[],
```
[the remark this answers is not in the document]
mkv-vcm: It's true, but the final form depends on what we can discuss about the mixing in of mint witnesses, so I'll leave this hanging for now.

src/interactions/signTx.ts
```diff
@@ -542,8 +615,13 @@ export function* signTransaction(version: Version, request: ParsedSigningRequest
...
     const isCatalystRegistrationSupported = getCompatibility(version).supportsCatalystRegistration
 
-    const witnessPaths = generateWitnessPaths(request)
-    const { tx, signingMode } = request
+    const { tx, signingMode, additionalWitnessPaths } = request
+    let witnessPaths: ValidBIP32Path[] = []
+    if (signingMode != TransactionSigningMode.MULTISIG_TRANSACTION) {
+        witnessPaths = generateWitnessPaths(request)
+    }
+    witnessPaths.concat(additionalWitnessPaths)
```
mkv-vcm, line 623: I wrote it like this, because this keeps the order of witnesses. If that's not important, we can spare a couple of lines and make it more const.

src/interactions/serialization/addressParams.ts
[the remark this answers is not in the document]
janmazak: I think we should get inspiration from other unsupported features, e.g.
```
    if (hasCatalystRegistration && !getCompatibility(version).supportsCatalystRegistration) {
        throw new DeviceVersionUnsupported(`Catalyst registration not supported by Ledger app version ${version}.`)
    }
```

src/parsing/address.ts
janmazak: This has to stay as it is to avoid integration problems (if someone used extensions of our types).

src/interactions/signTx.ts
janmazak: Hm. I would say it makes sense to make witnesses unique?... By removing duplicates from the additional requests. I don't see any harm in that.

src/interactions/serialization/addressParams.ts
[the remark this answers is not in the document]
mkv-vcm: ensureRequestSupportedByAppVersion is not called from derive and show address. But yeah, I need to extend the checks in that function as well.

src/interactions/signTx.ts
mkv-vcm: Done. Scrambles up the ordering, but we don't care about that apparently.

src/types/public.ts
```diff
@@ -1076,5 +1129,7 @@ export type SignTransactionRequest = {
      * Transaction to be signed
      */
     tx: Transaction
+
+    scriptWitnessPaths: BIP32Path[]
```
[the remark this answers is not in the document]
mkv-vcm: Yup, forgot between renamings.

src/interactions/signTx.ts
```diff
@@ -589,22 +589,48 @@ function ensureRequestSupportedByAppVersion(version: Version, request: ParsedSig
...
+        {
+            const scriptAddressTypes = [
+                AddressType.BASE_PAYMENT_KEY_STAKE_SCRIPT,
+                AddressType.BASE_PAYMENT_SCRIPT_STAKE_KEY,
+                AddressType.BASE_PAYMENT_SCRIPT_STAKE_SCRIPT,
+                AddressType.ENTERPRISE_SCRIPT,
+                AddressType.POINTER_SCRIPT,
+                AddressType.REWARD_SCRIPT,
+            ]
+            if (request?.tx?.outputs && request.tx.outputs.some(o =>
```
janmazak, line 614: Maybe we can pull the complicated condition outside of the `if`, as we do with `hasScripthashStakeCredentials`.

src/interactions/signTx.ts
```diff
@@ -616,11 +642,12 @@ export function* signTransaction(version: Version, request: ParsedSigningRequest
...
     const { tx, signingMode, additionalWitnessPaths } = request
-    let witnessPaths: ValidBIP32Path[] = []
+    let witnessPaths = additionalWitnessPaths
     if (signingMode != TransactionSigningMode.SCRIPT_TRANSACTION) {
-        witnessPaths = generateWitnessPaths(request)
+        witnessPaths = witnessPaths.concat(generateWitnessPaths(request))
     }
-    witnessPaths = witnessPaths.concat(additionalWitnessPaths)
+    // throw out repeated paths
+    witnessPaths = Array.from(new Set(witnessPaths))
```
janmazak, line 650: I am quite wary of relying on internal workings of the Set type --- how are paths tested for equality? I don't really know, it has apparently changed over time (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), and so I would prefer an explicit approach as we have used in `generateWitnessPaths`. Btw. the logic is not needed there anymore, so it should just be extracted into a function and moved here.

src/interactions/signTx.ts
mkv-vcm: We certainly can!

src/interactions/signTx.ts
mkv-vcm: You are right, and this shows my lack of practice with JS. It's a weird language. I'm not sure I understand how the Record type works (type Record<K extends keyof any, T>), but I guess we can trust that it does, and so I can use that.

janmazak: Yes, while I don't know much about Record, the existing approach was written by @ppershing and reviewed by @refi93, so I am pretty sure I can't come up with anything better than them.

mkv-vcm: Done. I assume it works based on the Object key equality, which is well defined for strings, and the JSON representation is supposedly unique for paths.

ppershing: Set checks for triple equals, which means referential equality for objects. Definitely not something you want here I believe

src/interactions/signTx.ts
```diff
@@ -627,12 +622,21 @@ function ensureRequestSupportedByAppVersion(version: Version, request: ParsedSig
...
-            throw new DeviceVersionUnsupported(`Scripthash based withdrawal not supported by Ledger app version ${version}.`)
+        {
+            const withdrawals = request?.tx?.withdrawals
+            const hasScripthashWithdrawals = withdrawals && withdrawals.some(w => w.stakeCredential.type == StakeCredentialType.SCRIPT_HASH)
+            if (hasScripthashWithdrawals) {
+                throw new DeviceVersionUnsupported(`Scripthash based withdrawal not supported by Ledger app version ${version}.`)
+            }
         }
-    }
+}
```
janmazak, line 632: Indentation?

mkv-vcm: Yes.

src/types/public.ts
```diff
@@ -637,14 +653,21 @@ export type PoolRetirementParams = {
     retirementEpoch: bigint_like,
 }
 
+export type StakeCredentialParams = {
+    path?: BIP32Path,
+    scriptHash?: string,
+}
```
mkv-vcm, line 659: We want a type distinction there, so that the type mechanism ensures exactly one of path/scriptHash is given.
See how Certificate is defined, it seems to me we can do it analogously to that.
  -- JM

src/interactions/serialization/txCertificate.ts
```diff
@@ -1,24 +1,62 @@
...
 
-export function serializeTxCertificate(
+export function serializeTxCertificatePreMultisig(
     certificate: ParsedCertificate,
 ) {
-
     switch (certificate.type) {
     case CertificateType.STAKE_REGISTRATION:
     case CertificateType.STAKE_DEREGISTRATION: {
+        validate(certificate.stakeCredential.type == StakeCredentialType.KEY_PATH, InvalidDataReason.CERTIFICATE_INVALID_STAKE_CREDENTIAL)
```
mkv-vcm, line 14: This validation must be done during parsing or in ensureRequestSupportedByAppVersion.
-- JM

src/interactions/serialization/txCertificate.ts
```diff
@@ -1,24 +1,62 @@
...
         return Buffer.concat([
             uint8_to_buf(certificate.type as Uint8_t),
-            path_to_buf(certificate.path),
+            path_to_buf(certificate.stakeCredential.path),
         ])
     }
     case CertificateType.STAKE_DELEGATION: {
+        validate(certificate.stakeCredential.type == StakeCredentialType.KEY_PATH, InvalidDataReason.CERTIFICATE_INVALID_STAKE_CREDENTIAL)
```
mkv-vcm, line 21: This validation must be done during parsing or in ensureRequestSupportedByAppVersion.
-- JM

src/interactions/serialization/txOther.ts
```diff
@@ -10,12 +13,22 @@ export function serializeTxInput(
     ])
 }
 
+export function serializeTxWithdrawalPreMultisig(
+    withdrawal: ParsedWithdrawal
+) {
+    validate(withdrawal.stakeCredential.type == StakeCredentialType.KEY_PATH, InvalidDataReason.WITHDRAWAL_INVALID_IDENTIFIER)
```
mkv-vcm, line 19: This validation must be done during parsing or in ensureRequestSupportedByAppVersion.
-- JM

src/utils/serialize.ts
```diff
@@ -46,7 +43,17 @@ export function buf_to_uint32(data: Buffer): Uint32_t {
...
+    return Buffer.concat([padding, data])
+}
+
+export function int64_to_buf(value: Int64_str): Buffer {
+    assert(isInt64str(value), 'invalid int64_str')
+
+    const data = new Int64BE(value, 10).toBuffer()
     assert(data.length <= 8, "excessive data")
 
     const padding = Buffer.alloc(8 - data.length)
```
mkv-vcm, line 59: toBuffer always returns 8 bytes, so padding is unnecessary. But will have to change the assert.
-- JM

src/types/public.ts
```diff
@@ -190,26 +204,25 @@ export type AddressParamsBase = {
  */
 // Not really worth the effort of disambiguation through additional tagged enum
```
mkv-vcm, line 205: I am not sure this is still true. I find the address params types messy. The staking part that goes into base address is the same as what goes into a reward address (maybe one of the options is forbidden in one of those places), but the types completely disregard that. We should get the opinion of Adalite's people who are integrating this stuff.
-- JM

mkv-vcm (commented):
From the #120 PR.

src/types/public.ts
mkv-vcm: Done.

src/interactions/serialization/txCertificate.ts
mkv-vcm: Done.

src/interactions/serialization/txCertificate.ts
mkv-vcm: Done.

src/interactions/serialization/txOther.ts
mkv-vcm: Can't validate it before knowing app version, and already present in ensure... so just turned into an assert.

src/utils/serialize.ts
mkv-vcm: Done.

src/types/public.ts
mkv-vcm: What is the procedure with this one @janmazak

src/types/public.ts
```diff
@@ -653,11 +654,23 @@ export type PoolRetirementParams = {
...
 
-export type StakeCredentialParams = {
-    path?: BIP32Path,
-    scriptHash?: string,
+export enum StakeCredentialParamsType {
+    KEY_PATH,
+    SCRIPT_HASH
 }
 
+export type KeyStakeCredential = {
```
janmazak, line 662: I'm not sure how it is done elsewhere with similar types, but don't we want to keep "params" in the name since it is a part of StakeCredentialParams?

src/interactions/serialization/addressParams.ts
```diff
@@ -1,53 +1,93 @@
...
 export function serializeAddressParams(
-    params: ParsedAddressParams
+    params: ParsedAddressParams,
+    version: Version,
 ): Buffer {
+    let spending: SpendingDataSource = params.spendingDataSource
+    let staking: StakingDataSource = params.stakingDataSource
+    if (!getCompatibility(version).supportsScriptTransaction) {
+        if (params.type == AddressType.REWARD_KEY) {
+            validate(staking.type == StakingDataSourceType.KEY_PATH, InvalidDataReason.ADDRESS_INVALID_STAKING_INFO)
```
janmazak, line 73: assert?

src/utils/serialize.ts
```diff
@@ -46,11 +43,19 @@ export function buf_to_uint32(data: Buffer): Uint32_t {
 export function uint64_to_buf(value: Uint64_str): Buffer {
     assert(isUint64str(value), 'invalid uint64_str')
 
-    const data = bs10.decode(value)
-    assert(data.length <= 8, "excessive data")
+    const data = new Uint64BE(value, 10).toBuffer()
+    assert(data.length == 8, "excessive data")
```
janmazak, line 47: "excessive data" does not capture the insufficient data case (also below), let's use "invalid data length" or something

src/types/public.ts
janmazak: I've created an issue for this, #126 . Unfortunately, it is part of the public interface, so I believe we should do some cleanup before the final release of ledgerjs. (But it does not block merging this PR.)

src/types/public.ts
mkv-vcm: Kinda agree, but it looks a bit weird.

src/interactions/serialization/addressParams.ts
mkv-vcm: Done.

src/utils/serialize.ts
mkv-vcm: True.

janmazak:
Will be merged via #130.

PR #129, opened 2021-08-30
Canonical ordering tests

mkv-vcm:
Tests that I hope cover most of the error cases with canonical ordering in place now.


src/parsing/transaction.ts
```diff
@@ -90,8 +90,8 @@ function parseTokenBundle<T>(tokenBundle: AssetGroup[], emptyTokenBundleAllowed:
     validate(policyIds.length == new Set(policyIds).size, InvalidDataReason.MULTIASSET_INVALID_TOKEN_BUNDLE_NOT_UNIQUE)
 
     // enforcing of policies order is removed for now and will be added back after the ordering is properly defined by a CIP
```
janmazak, line 92: This comment should go away.

src/parsing/transaction.ts
```diff
@@ -71,11 +71,11 @@ function parseAssetGroup<T>(assetGroup: AssetGroup, parseTokenAmountFn: ParseTok
     validate(assetNamesHex.length == new Set(assetNamesHex).size, InvalidDataReason.MULTIASSET_INVALID_ASSET_GROUP_NOT_UNIQUE)
 
     // enforcing of asset order is removed for now and will be added back after the ordering is properly defined by a CIP
```
janmazak, line 73: to be removed.

src/parsing/transaction.ts
```diff
@@ -90,8 +90,8 @@ function parseTokenBundle<T>(tokenBundle: AssetGroup[], emptyTokenBundleAllowed:
     validate(policyIds.length == new Set(policyIds).size, InvalidDataReason.MULTIASSET_INVALID_TOKEN_BUNDLE_NOT_UNIQUE)
```
janmazak, line 90: I think we do not need this anymore?

src/parsing/transaction.ts
```diff
@@ -71,11 +71,11 @@ function parseAssetGroup<T>(assetGroup: AssetGroup, parseTokenAmountFn: ParseTok
     validate(assetNamesHex.length == new Set(assetNamesHex).size, InvalidDataReason.MULTIASSET_INVALID_ASSET_GROUP_NOT_UNIQUE)
```
janmazak, line 71: Replaced by the ordering requirement? (I don't think the error of duplicit entries will be common, so we can get rid of the MULTIASSET_INVALID_ASSET_GROUP_NOT_UNIQUE exception and reduce validation to just the ordering.

src/parsing/transaction.ts
```diff
@@ -119,6 +119,7 @@ export function parseTransaction(tx: Transaction): ParsedTransaction {
     const certificates = parseCertificates(tx.certificates ?? [])
 
     // withdrawals
+    // we can't check here, but withdrawal keys (the StakeCredential) should be in canonical ordering
```
janmazak, line 122:
```suggestion
    // we can't check here, but withdrawal map keys (derived from stake credentials) should be in CBOR canonical ordering
```

janmazak: The comments on "Transaction" should be updated (i.e. in all three places where we require canonical ordering). See
```
    /**
     * Mint or burn instructions (if any).
     * Assets to be minted (token amount positive) or burned (token amount negative) with the transaction.
     * If not null, the entries' keys (policyIds) must be sorted lowest value to highest to reflect a canonical CBOR.
     * The sorting rules (as described in the [CBOR RFC](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9)) are:
     *  * if two keys have different lengths, the shorter one sorts earlier;
     *  * if two keys have the same length, the one with the lower value in lexical order sorts earlier.
     */
    // TODO update the docstring when CIP-0021 is approved
    mint?: Array<AssetGroup> | null,
```

src/parsing/transaction.ts
mkv-vcm: What would the advantage of that be? Reducing validation. The sorting doesn't solve this one.

src/parsing/transaction.ts
janmazak: I think it's OK not to reference a particular CIP for that (it does not have a definite number yet), just CBOR canonical RFC is OK, we probably don't need to copy its contents here.

src/parsing/transaction.ts
mkv-vcm: Yeah.

src/parsing/transaction.ts
mkv-vcm: Exactly.

src/parsing/transaction.ts
janmazak: Ah, I see, this version of validation via sorting (unlike the one in ledger) does not exclude duplicity. OK, let's keep it, then.

janmazak (approved):

janmazak:
Merged into #130.

PR #130, opened 2021-08-31
Multisig final

janmazak:


src/types/public.ts
```diff
@@ -1076,5 +1149,182 @@ export type SignTransactionRequest = {
...
+ * Defines in what format is the resulting native script hash shown on the
+ * Ledger device
+ * @category Scripts
+ * @see [[DeriveNativeScriptHashRequest]]
+ */
+export enum NativeScriptHashDisplayFormat {
+    /**
+     * The native script hash will be shown in a bech32 encoded format
+     */
+    BECH32 = 'bech32',
```
gabrielKerekes, line 1324: On Trezor I also added a `HIDE` option for when the script hash should only be derived without showing the script hash to the user. I'm not sure we need to include it now, but if it makes sense to have a `HIDE` option we should at least create an issue for it so it gets added at some point.

src/Ada.ts
```diff
@@ -314,10 +318,34 @@ export class Ada {
...
       const version = yield* getVersion()
       return yield* signOperationalCertificate(version, request)
   }
+
+  /**
+   * Derive a native script hash for the specified native script and display
+   * it on Ledger in the specified format. The hash is returned in raw hex
+   * format without any encoding.
+   */
+  async deriveNativeScriptHash(
```
gabrielKerekes, line 332: I just realised that `deriveNativeScriptHash` is not an ideal name for this operation. It's more `getNativeScriptHash` than `derive`. Addresses really are derived based on a path but here the hash is simply calculated and returned. On Trezor the operation is called `getNativeScriptHash`.

gabrielKerekes (commented):
I just found two things I should have noticed earlier, sorry.

src/types/public.ts
```diff
@@ -1056,6 +1113,22 @@ export enum TransactionSigningMode {
...
+    /**
+     * Represents a transaction controlled by scripts.
+     *
+     * Like an ordinary transaction, but stake credentials and all similar elements are given as script hashes
+     * and witnesses are decoupled from transaction elements.
+     *
+     * The transaction
+     * - *must* have `path=null` on all `inputs`
+     * - *must not* contain a pool registration certificate
+     * TODO
```
janmazak, line 1126: Must be resolved.

src/types/public.ts
```diff
@@ -1076,5 +1149,182 @@ export type SignTransactionRequest = {
      * Transaction to be signed
      */
     tx: Transaction
+
+    additionalWitnessPaths: BIP32Path[]
```
janmazak, line 1153: Must be documented.

janmazak: And should be made optional.

src/parsing/transaction.ts
```diff
@@ -215,14 +226,74 @@ export function parseSigningMode(mode: TransactionSigningMode): TransactionSigni
 export function parseSignTransactionRequest(request: SignTransactionRequest): ParsedSigningRequest {
     const tx = parseTransaction(request.tx)
     const signingMode = parseSigningMode(request.signingMode)
+    const additionalWitnessPaths = request.additionalWitnessPaths ? request.additionalWitnessPaths.map(path => parseBIP32Path(path, InvalidDataReason.INVALID_PATH)) : []
```
KubqoA, line 229: I think this could be written also like this:
```suggestion
    const additionalWitnessPaths = request.additionalWitnessPaths?.map(path => parseBIP32Path(path, InvalidDataReason.INVALID_PATH)) || []
```

PR #132, opened 2021-09-22
Alonzo phase 1

mkv-vcm:


src/types/public.ts
```diff
@@ -313,6 +313,10 @@ export type TxOutput = {
      * Destination address of the output
      */
     destination: TxOutputDestination
```
janmazak, line 315: I'd move this to be the first item in the object.

src/types/public.ts
```diff
@@ -313,6 +313,10 @@ export type TxOutput = {
...
+    /**
+     * Optional script data hash
+     */
+    scriptDataHash?: string | null
```
janmazak, line 319: I am against using "script" in this name, CDDL does not do it and formal specs also talk only about datum, no mention of scripts.

src/types/public.ts
```diff
@@ -1053,6 +1061,10 @@ export type Transaction = {
      * i.e. the key with the lower value in lexical order sorts earlier.
      */
     mint?: Array<AssetGroup> | null,
+    /**
+     * Script Data hash (if any)
+     */
+    scriptDataHash?: string | null,
```
janmazak, line 1067: Usually, we use "hex" in the names of hex strings, e.g.
```
export type TxAuxiliaryDataArbitraryHashParams = {
    /** Hash of the transaction auxiliary data */
    hashHex: string
}
```

janmazak (changes requested):
Just a few quick comments before I forget them.

src/types/public.ts
mkv-vcm: True, that explains the name collision misunderstanding in the Ledger code as well.

src/types/public.ts
mkv-vcm: Done.

src/types/public.ts
mkv-vcm: Refactoring, but yeah, it was weird.

src/interactions/signTx.ts
```diff
@@ -93,6 +95,15 @@ function* signTx_addOutput(
 
   yield* signTx_addTokenBundle(output.tokenBundle, P1.STAGE_OUTPUTS, uint64_to_buf)
 
+  if (output.datumHashHex) {
```
janmazak, line 98:
```suggestion
  if (output.datumHashHex != null) {
```

janmazak: (to be consistent with other places like this, it distinguishes undefined and null from anything else; that way, we do not need to rely on validation to catch stuff like empty arrays)

src/interactions/signTx.ts
```diff
@@ -80,6 +81,7 @@ function* signTx_addOutput(
 ): Interaction<void> {
   const enum P2 {
     BASIC_DATA = 0x30,
+    SCRIPT_DATA_HASH = 0x34,
```
janmazak, line 84:
```suggestion
    DATUM_HASH = 0x34,
```

src/parsing/transaction.ts
```diff
@@ -148,6 +152,7 @@ export function parseTransaction(tx: Transaction): ParsedTransaction {
         certificates,
         fee,
         mint,
+        scriptDataHashHex: scriptDataHash,
```
janmazak, line 155:
```suggestion
        scriptDataHashHex,
```

src/parsing/transaction.ts
```diff
@@ -137,6 +137,10 @@ export function parseTransaction(tx: Transaction): ParsedTransaction {
         ? null
         : parseTokenBundle(tx.mint, false, parseInt64_str)
 
+    const scriptDataHash = tx.scriptDataHashHex == null
```
janmazak, line 140:
```suggestion
    const scriptDataHashHex = tx.scriptDataHashHex == null
```

src/parsing/transaction.ts
```diff
@@ -205,10 +210,16 @@ function parseTxOutput(
...
+
+    const datumHash = output.datumHashHex
+        ? parseHexStringOfLength(output.datumHashHex, SCRIPT_DATA_HASH_LENGTH, InvalidDataReason.SCRIPT_DATA_HASH_WRONG_LENGTH)
+        : null
+
     return {
         amount,
         tokenBundle,
         destination,
+        datumHashHex: datumHash,
```
janmazak, line 222:
```suggestion
        datumHashHex,
```

src/parsing/transaction.ts
```diff
@@ -205,10 +210,16 @@ function parseTxOutput(
     const tokenBundle = parseTokenBundle(output.tokenBundle ?? [], true, parseUint64_str)
 
     const destination = parseTxDestination(network, output.destination)
+
+    const datumHash = output.datumHashHex
```
janmazak, line 214:
```suggestion
    const datumHashHex = output.datumHashHex
```

janmazak: Also, let's write the ?: operator consitently with other cases, e.g.
```
const ttl = tx.ttl == null
        ? null
        : parseUint64_str(tx.ttl, {}, InvalidDataReason.TTL_INVALID)
```

test/integration/__fixtures__/signTx.ts
```diff
@@ -523,12 +537,18 @@ export const outputs: Record<
         ],
     },
     datumHash: {
-        destination: destinations.internalBaseWithStakingPath,
+        destination: destinations.internalBaseWithStakingScript,
+        amount: 7120787,
+        datumHashHex: "ffd4d009f554ba4fd8ed1f1d703244819861a9d34fd4753bcf3ff32f043ce188",
+    },
+    datumHashExternal: {
+        // destination: destinations.multiassetThirdParty,
```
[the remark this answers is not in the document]
janmazak: Can't the names be improved? They can be as long as we need to make them clear, hopefully without additional comments.

test/integration/__fixtures__/signTx.ts
```diff
@@ -618,7 +675,7 @@ export type TestcaseShelley = {
   testname: string
   tx: Transaction
   signingMode: TransactionSigningMode
-    additionalWitnessPaths: BIP32Path[],
+  additionalWitnessPaths: BIP32Path[]
```
janmazak, line 678: The trailing comma is not consistent here --- don't we have some eslint rule for that, actually? Or is it not applied here? (Perhaps consult with @KubqoA how to deal with it in some automated fashion.)

test/integration/__fixtures__/signTx.ts
```diff
@@ -680,14 +738,15 @@ export const testsShelleyNoCertificates: TestcaseShelley[] = [
         },
         signingMode: TransactionSigningMode.MULTISIG_TRANSACTION,
         additionalWitnessPaths: [str_to_path("1854'/1815'/0'/0/0")],
+        txBody: "a500818258203b40265111d8bb3c3c608d95b3a0bf83461ace32d79336579a1939b3aad1c0b700018182583d105e2f080eb93bad86d401545e0ce5f2221096d6477e11e6643922fa8d2ed495234dc0d667c1316ff84e572310e265edb31330448b36b7179e28dd419e0102182a030a0f00",
         result: {
             txHashHex:
-        "23d82edc8fbd2d55237cba955a2280161ebd5643b23844e9b5abdc843b966e62",
+        "d2c7763e9610c3440a8000cb6abfcf3f48d4ef713505066cea901a4ef576db16",
```
janmazak, line 744: I think we should add a test checking for consistency: for each test, validate that txHashHex is correctly computed from txBody (perhaps at the beginning of running a test suite? that way, it will force us to update it, hopefully). Or just write the tests in such a way that that the correct result is computed from the given tx body? But that would be a bit annoying for debugging (one wants a quick way to manually compare the tx hash with ledger result).

Perhaps create a separate issue about this.

janmazak (changes requested):

test/integration/__fixtures__/signTx.ts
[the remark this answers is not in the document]
mkv-vcm: I guess they could be, though I'm not sure how to gauge exactly what is in the hashes.

src/errors/invalidDataReason.ts
```diff
@@ -132,6 +132,8 @@ export enum InvalidDataReason {
 
   VALIDITY_INTERVAL_START_INVALID = "invalid validity interval start",
 
+  SCRIPT_DATA_HASH_WRONG_LENGTH = "script data hash not 32 long",
```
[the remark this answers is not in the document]
mkv-vcm: Reminder for myself that there is another similar one in second phase.

src/interactions/signTx.ts
```diff
@@ -31,6 +31,7 @@ const enum P1 {
   STAGE_VALIDITY_INTERVAL_START = 0x09,
   STAGE_CONFIRM = 0x0a,
   STAGE_MINT = 0x0b,
+  STAGE_SCRIPT_DATA_HASH = 0x0c,
```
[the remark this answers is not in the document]
mkv-vcm: That can be arranged.

src/interactions/signTx.ts
mkv-vcm: Hmm, I'm not a hundred percent sure what the exact reasoning here is. One catches anything that is evaluates to false, including null and undefined, and the latter will only refuse undefined and null. What validations don't we want to rely on here? That member is already a FixlenHexString.

src/interactions/signTx.ts
mkv-vcm: Done.

src/parsing/transaction.ts
mkv-vcm: With one more line.

src/parsing/transaction.ts
mkv-vcm: Ok.

test/integration/__fixtures__/signTx.ts
mkv-vcm: That's a neat idea for automation, I'm just a bit worried about bringing the hash calculating code into the JS side. It probably won't change, and if it does it's a huge breaking change for the whole infrastructure, but still, that feels like an internal thing. But definitely worth a look at.

mkv-vcm: #142

test/integration/__fixtures__/signTx.ts
mkv-vcm: In progress with Jakub.

janmazak (commented):
There are some unresolved comments here, e.g. the one about "in progress with Jakub".

test/integration/__fixtures__/signTx.ts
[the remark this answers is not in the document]
mkv-vcm: Yeah, that'll be it I think, I can't really explain the hashes.

src/types/public.ts
```diff
@@ -934,6 +934,11 @@ export type SignedTransactionData = {
      * the caller with information needed to assemble the transation containing these auxiliary data.
      */
     auxiliaryDataSupplement: TxAuxiliaryDataSupplement | null,
+    /**
+     * A flag to note if network id is part of the tx body, since that cannot be detected any other way,
+     * and the exact same transaction creating different tx bodies and hashes might cause problems
```
janmazak, line 939: Not entirely true, I'd use something like:
A flag indicating whether network id is part of the serialized tx body --- this depends on Ledger app version, not on the transaction itself (or other parameters of the signTransaction() call).

janmazak: Also, we should make this flag optional, and comment that if not given, it defaults to false (for backwards compatibility) --- please consult with David to make sure we are consistent with Trezor.

This will render `NetworkIdlessTestResult` unnecessary perhaps?

janmazak:
Will be merged via #136.

PR #143, opened 2021-11-03
tests for single account model

mkv-vcm:


janmazak (approved):
CircleCI fails on `yarn lint`, otherwise OK.

gabrielKerekes:
I think it would be worth it to add all the tests which Trezor has so that we can be sure that all the combinations work as expected -> https://github.com/trezor/trezor-firmware/blob/master/common/tests/fixtures/cardano/sign_tx.failed.json#L1585. But maybe it can be considered an overkill for Ledger? What do you think @janmazak?

janmazak (changes requested):
I've added a TODO describing what needs to be added.

janmazak (approved):

PR #150, opened 2022-01-26
Fix shelljs vulnerability

janmazak:
Fixes minor issues reported by `yarn audit`, and some other minor errors/warnings.

WARNING: Integration tests not run yet, do not merge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "sinon": "^11.1.2",
     "ts-mock-imports": "^1.3.7",
     "ts-node": "^9.1.1",
-    "typedoc": "git+https://github.com/ppershing/typedoc.git#do_not_sort_properties",
+    "typedoc": "^0.22.11",
     "typescript": "^4.1.5"
   },
   "bugs": {
@@ -65,7 +65,6 @@
   },
   "homepage": "https://github.com/cardano-foundation/ledgerjs-hw-app-cardano",
   "scripts": {
-    "prepare": "cd node_modules/typedoc && npm install && npm run build",
     "prepack": "yarn build:flowtypes && yarn flow",
     "lint": "yarn eslint . --max-warnings=0",
     "clean": "rm -rf dist/",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -86,6 +84,8 @@
     "prettier-eslint/**/ansi-regex": "<6.0.0",
     "eslint/**/ansi-regex": "<6.0.0",
     "@ledgerhq/hw-transport-node-hid/**/ansi-regex": "<6.0.0",
-    "mocha/**/ansi-regex": "<6.0.0"
+    "mocha/**/ansi-regex": "<6.0.0",
+    "@ledgerhq/hw-transport-node-hid/**/simple-get": "^4.0.1",
+    "flowgen/**/shelljs": "^0.8.5"
```
refi93, line 89: I think this resolution isn't needed - tried removing it and probably the version of shelljs installed satisfied flowgen's (or its recursive dependencies) requirements

package.json
```diff
@@ -86,6 +84,8 @@
     "prettier-eslint/**/ansi-regex": "<6.0.0",
     "eslint/**/ansi-regex": "<6.0.0",
     "@ledgerhq/hw-transport-node-hid/**/ansi-regex": "<6.0.0",
-    "mocha/**/ansi-regex": "<6.0.0"
+    "mocha/**/ansi-regex": "<6.0.0",
+    "@ledgerhq/hw-transport-node-hid/**/simple-get": "^4.0.1",
```
refi93, line 88: apparently simple-get meanwhile got out a patched version 3.1.1 so this resolution isn't needed anymore either

package.json
janmazak: It is here to avoid shelljs 0.8.4 (which would get installed otherwise, at least it did when I tried). Do you mean that when I put a proper version into yarn.lock, I don't need it in resolutions anymore?

refi93: you probably just ran `yarn install` which doesn't try to update packages unless strictly required by some of the packages (shelljs version requirement `^0.8.4` in flowgen's package.json is satisfied both by `0.8.4.` and `0.8.5`

https://classic.yarnpkg.com/en/docs/dependency-versions/

refi93 (approved):

PR #167, opened 2022-07-25
Item 18 reference inputs

rustamlabs:


src/interactions/serialization/txOther.ts
```diff
@@ -109,3 +109,12 @@ export function serializeTotalCollateral(
         uint64_to_buf(totalCollateral),
     ])
 }
+
+export function serializeReferenceInputs(
```
janmazak, line 113:
```suggestion
export function serializeReferenceInput(
```

janmazak: I don't see a function to serialize collateral inputs. But it seems preferable to me to only have a single function for serializing ParsedInput used in 3 places --- that's how the APDU parsing side is written, anyway.

src/interactions/signTx.ts
```diff
@@ -623,6 +624,20 @@ function* signTx_addTotalCollateral(
     })
 }
 
+function* signTx_addTotalReferenceInputs(
+    referenceInputs: ParsedInput
```
janmazak, line 628: This does not seem right --- why is the type only a single input?

rustamlabs: this method called for every input one by on, similar to how  add collateral  is being added

src/interactions/signTx.ts
```diff
@@ -623,6 +624,20 @@ function* signTx_addTotalCollateral(
     })
 }
 
+function* signTx_addReferenceInput(
+    referenceInputs: ParsedInput
```
janmazak, line 628:
```suggestion
    referenceInput: ParsedInput
```

src/interactions/signTx.ts
```diff
@@ -971,6 +986,13 @@ export function* signTransaction(version: Version, request: ParsedSigningRequest
         yield* signTx_addTotalCollateral(tx.totalCollateral)
     }
 
+    // totalCollateral
```
janmazak, line 989: misplaced comment

test/integration/__fixtures__/signTxPlutus.ts
```diff
@@ -466,4 +466,52 @@ export const testsBabbage: SignTxTestcase[] = [
             auxiliaryDataSupplement: null,
         },
     },
+    {
+        // tx does not contain any Plutus elements, but should be accepted (differs only in UI)
+        testname: "Sign tx with change output (Plutus) and Reference Input MAP",
```
janmazak, line 471: What does "MAP" mean here? We use arrays for inputs.

janmazak (commented):
Take also a look at reject tests. They more-or-less copy security policies --- e.g. if you added some restrictions on tx signing mode in security policies for reference inputs, we need
1. add the proper validation to ledgerjs
2. add reject tests to ledgerjs checking the validation works the same in both ledgerapp and ledgerjs

janmazak (approved):

PR #168, opened 2022-07-27
Output inline datum

rustamlabs:


src/interactions/serialization/txOutput.ts
```diff
@@ -38,16 +44,75 @@ export function serializeTxOutputBasicParams(
     output: ParsedOutput,
     version: Version,
 ): Buffer {
-    const hasDatumHex = (output.type === TxOutputType.MAP_BABBAGE && output.datum?.type === DatumType.HASH) ||(output.type !== TxOutputType.MAP_BABBAGE && output.datumHashHex != null)
-    const datumHashHexBuffer = getCompatibility(version).supportsAlonzo
-        ? serializeOptionFlag(hasDatumHex)
+    const hasDatumOption = output.type === TxOutputType.MAP_BABBAGE
```
janmazak, line 47:
```suggestion
    const hasDatum = output.type === TxOutputType.MAP_BABBAGE
```

src/interactions/serialization/txOutput.ts
```diff
@@ -38,16 +44,75 @@ export function serializeTxOutputBasicParams(
...
+        ? output.datum != null
+        : output.datumHashHex != null
+
+    const datumOptionBuffer = getCompatibility(version).supportsAlonzo
+        ? serializeOptionFlag(hasDatumOption)
         : Buffer.from([])
 
+    const serializationFormat = parseUint8_t(output.type === undefined
+        ? TxOutputType.ARRAY_LEGACY
+        : output.type, InvalidDataReason.OUTPUT_INVALID_FORMAT)
```
janmazak, line 57: This is very convoluted, I am not even sure I understand what it does. It does not make sense to me to use any parsing function here, `output` is already parsed (and as such, should have a type different from `undefined`). Perhaps we want a switch statement distinguishing on type?

src/interactions/serialization/txOutput.ts
```diff
@@ -38,16 +44,75 @@ export function serializeTxOutputBasicParams(
...
+    //TODO: maybe null for unsupported version?
+
     return Buffer.concat([
-        uint8_to_buf(parseUint8_t(output.type === undefined? TxOutputType.ARRAY_LEGACY : output.type, InvalidDataReason.OUTPUT_INVALID_FORMAT)),
+        uint8_to_buf(serializationFormat),
```
janmazak, line 61: This item should only be included if Ledger app version is Babbage. (that should be determined by something like `getCompatibility(version).supportsBabbage`)

src/interactions/serialization/txOutput.ts
```diff
@@ -38,16 +44,75 @@ export function serializeTxOutputBasicParams(
...
+                hex_to_buf(datumHex),
+            ])
+        }
+
+        default:
+            return Buffer.concat([])
+        }
+
+    } else {    //  Alonzo Format
+        const datumHashHex = parseHexStringOfLength(output.datumHashHex, DATUM_HASH_LENGTH, InvalidDataReason.OUTPUT_INVALID_DATUM_HASH)
```
janmazak, line 111: Parsing does not belong here. Take a look at where it is done in `parseTxOutput` on the `develop` branch.

janmazak (commented):
I did not look at everything yet.

But for tests, it is reasonable to minimize irrelevant clicking. Try to turn on the expert mode (in the ledger app menu) and see that in some of the tests you created, there are some useless warnings in the beginning (that could be removed by adding/modifying data in test fixtures).

src/interactions/serialization/txOutput.ts
rustamlabs: hmm, hat right, previously ParsedOutput also had optional property (type?: TxOutputType.ARRAY_LEGACY) but since i tried to fix issue with circleCI it has been changed. However that change is not reflected to everywhere. I will take a look on type distinctions for parsed outputs

src/interactions/serialization/txOutput.ts
rustamlabs: Let's do compatibility checks and validations in another PR. This branch does not include CollRet and is not compatible with latest jm/signtx branch.

PR #169, opened 2022-07-28
Item 16 collateral return and inline datums 

rustamlabs:


src/interactions/signTx.ts
```diff
@@ -600,6 +602,33 @@ function* signTx_addRequiredSigner(
...
+        data: serializeTxOutputBasicParams(collateralOutput, version),
+        expectedResponseLength: 0,
+    })
+
+    yield* signTx_addTokenBundle(collateralOutput.tokenBundle, P1.STAGE_COLLATERAL_OUTPUT, uint64_to_buf)
+
+    yield send({
+        p1: P1.STAGE_COLLATERAL_OUTPUT,
+        p2: P2.CONFIRM,
+        data: Buffer.alloc(0),
```
janmazak, line 627: This is different from the other PR (Buffer.from([])). I don't remember which way is better (there was some problem with Buffer constructors in node.js), but let's use the one used elsewhere in the code.

src/parsing/transaction.ts
```diff
@@ -174,26 +174,35 @@ export function parseTransaction(tx: Transaction): ParsedTransaction {
         ? null
         : parseTokenBundle(tx.mint, false, parseInt64_str)
 
+    // script data hex
```
janmazak, line 177:
```suggestion
    // script data hash hex
```

src/parsing/transaction.ts
```diff
@@ -174,26 +174,35 @@ export function parseTransaction(tx: Transaction): ParsedTransaction {
...
+        : parseTxOutput(tx.collateralOutput, tx.network)
+
     //total collateral
     const totalCollateral = tx.totalCollateral == null
         ? null
         : parseUint64_str(tx.totalCollateral, {max: MAX_LOVELACE_SUPPLY_STR}, InvalidDataReason.TOTAL_COLLATERAL_NOT_VALID)
 
-    //reference Inputs
+    //reference inputs
     const referenceInputs = tx.referenceInputs == undefined
```
janmazak, line 206:
```suggestion
    const referenceInputs = tx.referenceInputs == null
```
As elsewhere.

janmazak (approved):
I've not checked the tests, but that can be done later, after merging it (when other tests are added and we can check if they cover everything).

PR #171, opened 2022-08-02
Script reference, Validations and compatibility 

rustamlabs:


src/interactions/serialization/txOutput.ts
```diff
@@ -1,4 +1,4 @@
-import {MAX_DATUM_CHUNK_SIZE} from "../../parsing/constants"
+import {MAX_CHUNK_SIZE} from "../../parsing/constants"
```
janmazak, line 1: This constant should probably be not in "parsing" since it is related to APDU serialization.

src/interactions/serialization/txOutput.ts
```diff
@@ -37,21 +37,27 @@ export function serializeTxOutputBasicParams(
     output: ParsedOutput,
     version: Version,
 ): Buffer {
+    const serializationFormatBuffer = getCompatibility(version).supportsBabbage
+        ? uint8_to_buf(output.type as Uint8_t)
+        : Buffer.from([])
+
     const hasDatum = output.datum != null
```
janmazak, line 44: Perhaps we don't need this variable?

src/interactions/serialization/txOutput.ts
```diff
@@ -105,3 +116,28 @@ export function serializeTxOutputDatum(
     }
 }
 
+export function serializeTxOutputScriptHex(
```
janmazak, line 119:
```suggestion
export function serializeTxOutputRefScript(
```
Definitely no "hex" in the name of this function.

src/interactions/serialization/txOutput.ts
```diff
@@ -105,3 +116,28 @@ export function serializeTxOutputDatum(
...
+    output: ParsedOutput,
```
janmazak, line 120: Similarly to datums, I'd not pass in the whole output.

src/interactions/signTx.ts
```diff
@@ -939,6 +978,18 @@ function ensureRequestSupportedByAppVersion(version: Version, request: ParsedSig
...
 
+    if (request.tx.collateralOutput && !getCompatibility(version).supportsBabbage) {
+        throw new DeviceVersionUnsupported(`Collateral output not supported by Ledger app version ${getVersionString(version)}.`)
+    }
+
+    if (request.tx.totalCollateral && !getCompatibility(version).supportsBabbage) {
+        throw new DeviceVersionUnsupported(`Total collateral not supported by Ledger app version ${getVersionString(version)}.`)
+    }
+
+    if (request.tx.referenceInputs.length !=0 && !getCompatibility(version).supportsBabbage) {
```
janmazak, line 989:
```suggestion
    if (request.tx.referenceInputs.length > 0 && !getCompatibility(version).supportsBabbage) {
```

janmazak: I'd use > instead of != 0 also above, it would be more consistent with the codebase of ledgerjs.

src/parsing/transaction.ts
```diff
@@ -156,12 +157,20 @@ function parseDatum(output: TxOutput): ParsedDatum | null {
...
             : {
                 type: DatumType.HASH,
-                datumHashHex: parseHexStringOfLength(output.datumHashHex, SCRIPT_DATA_HASH_LENGTH, InvalidDataReason.SCRIPT_DATA_HASH_WRONG_LENGTH),
+                datumHashHex: parseHexStringOfLength(output.datumHashHex, SCRIPT_DATA_HASH_LENGTH, InvalidDataReason.OUTPUT_INVALID_DATUM_HASH),
             }
     }
     return datum
 }
 
+function parseScriptHex(output: TxOutput): HexString | null {
```
janmazak, line 166: We should not care about output type here, just parse the hex string, no need to take the whole output as argument. (This functions is not "parseAndValidateIfConsistentWithOutput".)

src/interactions/serialization/txOutput.ts
rustamlabs: moved to internal

src/parsing/transaction.ts
rustamlabs: I can carry it to parseTxOutput but one way or another we will need to check type before checking if scriptHex exists

PR #172, opened 2022-08-02
Tests alonzo trezor comparison

rustamlabs:


test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+            // "protocol_magic": 764824073,
+            // "network_id": 1,
+            // if Networks.Mainnet differs, the test should just explicitly give these
+            network: Networks.Mainnet,
+            inputs: [inputs.utxoMultisig],
+            outputs: [outputs.trezorParityBabbage, outputs.trezorParityDatumHashBabbage],
+            fee: 42,
+            ttl: 10,
+            validityIntervalStart: 47,
+            certificates: [
```
gabrielKerekes, line 1158: I think we can safely leave out the certificates, withdrawals, minting and auxiliary data since they are already tested in `Full test for trezor feature parity`.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+                },
+            ],
+            auxiliaryData: {
+                type: TxAuxiliaryDataType.ARBITRARY_HASH,
+                params: {
+                    hashHex: "58ec01578fcdfdc376f09631a7b2adc608eaf57e3720484c7ff37c13cff90fdf",
+                },
+            },
+            scriptDataHashHex: "3b40265111d8bb3c3c608d95b3a0bf83461ace32d79336579a1939b3aad1c0b7",
+            collateralInputs: [inputs.utxoByron],
```
gabrielKerekes, line 1219: I'd prefer if we used `utxoShelley` here - it doesn't make much difference but I wouldn't use Byron elements if we're not explicitly testing for them and in this case I don't think we need to be testing for them.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+            includeNetworkId: true,
+            collateralOutput: outputs.externalByronMainnet,
+            totalCollateral: 10,
+            referenceInputs: [inputs.utxoByron],
```
gabrielKerekes, line 1223: Same as above - I'd prefer `utxoShelley`.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+            auxiliaryData: {
+                type: TxAuxiliaryDataType.ARBITRARY_HASH,
+                params: {
+                    hashHex: "58ec01578fcdfdc376f09631a7b2adc608eaf57e3720484c7ff37c13cff90fdf",
+                },
+            },
+            scriptDataHashHex: "3b40265111d8bb3c3c608d95b3a0bf83461ace32d79336579a1939b3aad1c0b7",
+            collateralInputs: [inputs.utxoByron],
+            includeNetworkId: true,
+            collateralOutput: outputs.externalByronMainnet,
```
gabrielKerekes, line 1221: Same as above, I'd avoid Byron elements and would use `externalShelleyBaseKeyhashKeyhash`.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+            totalCollateral: 10,
+            referenceInputs: [inputs.utxoByron],
+
+        },
+        signingMode: TransactionSigningMode.PLUTUS_TRANSACTION,
+        additionalWitnessPaths: [str_to_path("1854'/1815'/0'/0/0"), str_to_path("1854'/1815'/0'/2/0")],
```
gabrielKerekes, line 1227: I don't think we need these after we remove minting.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+            auxiliaryData: {
+                type: TxAuxiliaryDataType.ARBITRARY_HASH,
+                params: {
+                    hashHex: "58ec01578fcdfdc376f09631a7b2adc608eaf57e3720484c7ff37c13cff90fdf",
+                },
+            },
+            scriptDataHashHex: "3b40265111d8bb3c3c608d95b3a0bf83461ace32d79336579a1939b3aad1c0b7",
+            includeNetworkId: true,
+        },
+        signingMode: TransactionSigningMode.MULTISIG_TRANSACTION,
```
gabrielKerekes, line 1323: This should be an `ORDINARY_TRANSACTION`.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+            // "protocol_magic": 764824073,
+            // "network_id": 1,
+            // if Networks.Mainnet differs, the test should just explicitly give these
+            network: Networks.Mainnet,
+            inputs: [inputs.utxoShelley],
+            outputs: [outputs.trezorParityBabbageOutputs],
+            fee: 42,
+            ttl: 10,
+            validityIntervalStart: 47,
+            certificates: [
```
gabrielKerekes, line 1260: Same as in the previous test - the extra elements can IMHO be removed - certificates, withdrawals, mint, auxiliary data and additional witnesses.

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
             auxiliaryDataSupplement: null,
         },
     },
+    {
+        testname: "Full test for trezor feature parity Babbage new elements",
```
gabrielKerekes, line 1147:
```suggestion
        testname: "Full test for trezor feature parity - Babbage elements (Plutus)",
```

test/integration/__fixtures__/signTx.ts
```diff
@@ -1143,4 +1143,199 @@ export const testsAlonzoTrezorComparison: SignTxTestcase[] = [
...
+                {
+                    path: str_to_path("1854'/1815'/0'/2/0"),
+                    witnessSignatureHex: "f64ce23e5c467775b64ffd7f18a2fda1dab92ddf2b6266d809bd014fd35f15a87d8451da7fd439b10ec763f9a2b120e520cc52abe0a92aa8295684f0b4c9f50d",
+                },
+            ],
+            auxiliaryDataSupplement: null,
+        },
+    },
+    {
+        testname: "Full test for trezor feature parity Babbage inline datum and script reference",
```
gabrielKerekes, line 1249:
```suggestion
        testname: "Full test for trezor feature parity - Babbage elements (ordinary)",
```

test/integration/__fixtures__/txElements.ts
```diff
@@ -733,6 +736,22 @@ export const outputs: Record<
             },
         ],
     },
+    trezorParityBabbage: {
+        type: TxOutputType.MAP_BABBAGE,
+        destination: destinations.multiassetThirdParty,
```
gabrielKerekes, line 741: I'd replace this with `destinations.externalShelleyBaseKeyhashKeyhash` just so the reader doesn't have to wonder, why `multiassetThirdParty` was used.

gabrielKerekes (approved):
In general the tests are ok as they are, I just think we don't need to include elements which are already tested in the previous Trezor parity test so we don't have to maintain those elements if they ever change. After you implement my suggestions I'll include the tests in Trezor.

PR #173, opened 2022-08-05
Babbage

janmazak:


src/types/public.ts
```diff
@@ -317,7 +336,48 @@ export type TxOutput = {
...
+    tokenBundle?: Array<AssetGroup> | null;
+    /**
+     * Optional datum hash or inline datum
+     */
+    datum?: Datum;
+    /**
+     * Optional reference script
+     * (without the #6.24 tag)
+     */
+    scriptHex?: string | null;
```
gabrielKerekes, line 372: Is there any difference between `scriptHex?: string | null` and `scriptHex?: string`? Why do we need the ` | null`?

gabrielKerekes: The question applies to all the places we use `optionalProp?: type | null` - e.g. in the `Transaction` type. If anything it seems a bit inconsistent since we add `| null` in some places and not in others. Is there some pattern?

src/types/public.ts
```diff
@@ -291,12 +291,31 @@ export type AssetGroup = {
...
- * Transaction output
- * @category Basic types
- * @see [[Transaction]]
+ * Corresponds to legacy_transaction_output in
+ * https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl
+ * Serializes as an array.
  */
-export type TxOutput = {
+export type TxOutputAlonzo = {
+    type?: TxOutputType.ARRAY_LEGACY;
```
gabrielKerekes, line 318: I noticed that `format` is used in the Ledger app. Wouldn't be better to use `format` and `TxOutputFormat` also in LedgerJS?

`format` also wouldn't be confused with `TxOutputDestination.type`. Right now outputs look like this: `{type: MAP_BABBAGE, destination: { type: THIRD_PARTY ... } ... }` i.e. there's two types.

What do you think?

src/types/public.ts
janmazak: https://stackoverflow.com/questions/57243477/can-an-optional-parameter-be-null-in-typescript

There is some meaning in x?: t | null, in particular, the argument can be left out, but if it is given, it cannot be undefined. I'm not saying we are using this correctly everywhere, though, I need to check.

janmazak: Should be fixed in b6fb7c94755c6dd92c55c224c054d48bddfedcf7.

gabrielKerekes (approved):

PR #176, opened 2022-08-25
fix: replace == with === in validation

janmazak:


src/parsing/transaction.ts
```diff
@@ -508,7 +508,7 @@ export function parseSignTransactionRequest(request: SignTransactionRequest): Pa
         // input should not be given with a path
         // the path is not used, but we check just to avoid potential confusion of developers using this
         validate(
-            tx.inputs.every(inp => inp.path == null),
+            tx.inputs.every(inp => inp.path === null),
```
gabrielKerekes, line 511: `null` checks should be left with only `==`. `a == null` is `true` if the value is `null` or `undefined` - which, unless I'm mistaken, is desired in these cases. `a === null` will result in `false` if `a` is `undefined`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined

janmazak: I've tried to make the comparisons consistent, e.g. we use `tx.totalCollateral === null` above.

I think these validations apply to tx of type `ParsedTransaction`, so all the undefineds should have been replaced by `null` during the parsing in `parseTransaction`. Am I wrong?

gabrielKerekes: Hm, I see 🤔 I think I'd prefer sticking to `==` in the `null` comparisons anyways as it seems to be safer if someone ever decided to change the `null`s to undefined for example or introduced another field which could be undefined and the comparisons would be copied without realising that they're incorrect.

But I'll leave it up to you.

janmazak: It's reasonable to use only ==. I've made it consistent across the function.

gabrielKerekes (approved):

PR #180, opened 2022-12-09
Transport (node v18)

janmazak:


src/Ada.ts
```diff
@@ -128,11 +128,12 @@ async function interact<T>(
  * @category Main
  */
 export class Ada {
-  /** @ignore */
+  /** $FlowIgnore[value-as-type] */
```
refi93, line 131: I§d leave here a comment claryfing why a flowignore comment is needed in an obviously ts codebase

PR #185, opened 2023-01-09
update node in circleci

janmazak:


janmazak:
Will be merged via #186.

PR #186, opened 2023-01-23
CIP36 renaming

janmazak:


test/integration/__fixtures__/signTxCVote.ts
```diff
@@ -75,28 +75,28 @@ export const testsCatalystRegistration: SignTxTestcase[] = [
...
             },
         },
     },
 ]
 
-export const testsGovernanceVotingRegistrationCIP36: SignTxTestcase[] = [
+export const testsCVoteRegistrationCIP36: SignTxTestcase[] = [
     {
-        testName: "Sign tx with governance voting registration CIP36 with voting key hex",
+        testName: "Sign tx with voting registration CIP36 with vote key hex",
```
davidmisiak, line 88: Also here (even though it's just a test case) - maybe we should use `vote registration` when the field is called `TxAuxiliaryDataType.CIP36_VOTE_REGISTRATION`.

src/interactions/signTx.ts
```diff
@@ -587,9 +587,9 @@ function* signTx_setAuxiliaryData(
     const signature = response.slice(AUXILIARY_DATA_HASH_LENGTH, AUXILIARY_DATA_HASH_LENGTH + ED25519_SIGNATURE_LENGTH)
 
     return {
-        type: TxAuxiliaryDataSupplementType.GOVERNANCE_VOTING_REGISTRATION,
+        type: TxAuxiliaryDataSupplementType.CIP36_VOTING_REGISTRATION,
```
davidmisiak, line 590: I think it should be `CIP36_VOTE_REGISTRATION`, the `TxAuxiliaryDataType` equivalent of this uses `VOTE` instead of `VOTING`.

janmazak: I've decided to just name it CIP36_REGISTRATION in both cases. (CIP36 does not actually mention "voting registration" except once in an example.)

src/interactions/signTx.ts
```diff
@@ -498,15 +498,15 @@ function* signTx_setAuxiliaryData(
...
   })
 
-  if (auxiliaryData.type === TxAuxiliaryDataType.GOVERNANCE_VOTING_REGISTRATION) {
+  if (auxiliaryData.type === TxAuxiliaryDataType.CIP36_VOTE_REGISTRATION) {
       const params = auxiliaryData.params
 
     const enum P2 {
       INIT = 0x36,
-      VOTING_KEY = 0x30,
+      VOTE_KEY = 0x30,
```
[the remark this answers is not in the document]
janmazak: It's a local enum and the context is totally unambiguous, so I'd rather keep it like this.

src/interactions/signTx.ts
davidmisiak: I agree. In that case also consider replacing `CIP36 voting registration` with `CIP36 registration` also in comments etc. (I know this is getting annoying, but it might be worth it considering the future confusion that could be caused by the left-over inconsistencies.)

davidmisiak (approved):

PR #187, opened 2023-01-31
update @ledgerhq/transport to include stax

janmazak:


xdzurman (approved):

package.json
```diff
@@ -21,16 +21,16 @@
   "author": "VacuumLabs <adalite@vacuumlabs.com>",
   "license": "Apache-2.0",
   "dependencies": {
-    "@ledgerhq/hw-transport": "^6.27.8",
+    "@ledgerhq/hw-transport": "^6.27.10",
     "@types/ledgerhq__hw-transport": "^4.21.3",
```
refi93, line 25: is this still needed? `@ledgerhq/hw-transport` seems to already have the types included

refi93 (approved):

package.json
janmazak: I don't know the exact reason, but I tried to remove it a while ago and it caused some error and I couldn't fix it quickly, so I gave up. Now removing the types does not cause any errors, so I'll try again.

PR #188, opened 2023-02-02
fix: number type for native script

janmazak:


davidmisiak (approved):
Note that there is `requiredCount?: bigint_like` also in `src/parsing/nativeScript.ts`.

PR #205, opened 2024-02-14
Message signing (CIP-8)

janmazak:


src/types/public.ts
```diff
@@ -294,6 +294,51 @@ export type OperationalCertificate = {
...
+ */
+export type MessageData = {
+  // message as bytes in hex
+  // Note: If the message is too long to be displayed at once, it must be hashed.
+  // MAX_CIP8_MSG_FIRST_CHUNK_ASCII_SIZE 198
+  // MAX_CIP8_MSG_FIRST_CHUNK_HEX_SIZE 99
+  messageHex: string
+
+  // the message will be signed with this key
+  witnessPath: BIP32Path
```
gabrielKerekes, line 323: I'd prefer `signingPath` here since `witness` is rather a transaction terminology and nowhere in [CIP08](https://cips.cardano.org/cip/CIP-0008) are witnesses mentioned.

I suppose the idea was to make it consistent with tx signing, but IMO these are two different concepts which don't need to be consistent.

src/types/public.ts
```diff
@@ -1276,6 +1325,17 @@ export type OperationalCertificateSignature = {
   signatureHex: string
 }
 
+/**
+ * CIP-8 message signature
+ * @category Basic types
+ * @see [[Ada.signMessage]]
+ */
+export type SignedMessageData = {
+  witnessSignatureHex: string
```
gabrielKerekes, line 1334: Same as above - I'd drop the witness terminology.

src/types/public.ts
```diff
@@ -1276,6 +1325,17 @@ export type OperationalCertificateSignature = {
...
+  witnessPath: BIP32Path
```
gabrielKerekes, line 1335: If public key is being returned I don't think we need to return the path as well - the path is passed in by the client and only a single signature is always returned. For tx signing the path is used mainly so the wallet is able to map the witness to a given key, no? So here we shouldn't need it?

src/types/public.ts
```diff
@@ -1276,6 +1325,17 @@ export type OperationalCertificateSignature = {
...
+  witnessPublicKeyHex: string
```
gabrielKerekes, line 1336:
```suggestion
  signingPublicKeyHex: string
```

or simply

```suggestion
  publicKeyHex: string
```

src/types/public.ts
```diff
@@ -1276,6 +1325,17 @@ export type OperationalCertificateSignature = {
   signatureHex: string
 }
 
+/**
+ * CIP-8 message signature
+ * @category Basic types
+ * @see [[Ada.signMessage]]
+ */
+export type SignedMessageData = {
```
gabrielKerekes, line 1333: I see that headers are not being returned here. Do you think it's not a good idea to return them?

I'm not sure tbh. Trezor does return them ([here](https://github.com/vacuumlabs/trezor-suite/pull/32/files#diff-cd1b26ff81881629a60750a5b48529d32141deb838b90dc8d802670a43f42b50R103)) but I just checked and it seems that Trezor FW does not return them they're only built by Connect. Which I'm not 100% sure is that useful. The wallet has the `address` and it also knows if `hashPayload` was set to true. The algorithm is a constant and I'm not sure about the `version` field. But maybe the wallet can build the headers by itself and so Trezor would not need to return them as well?

Opinions? cc Peto (whom I can't tag here, I'll tag him on slack... 🙈 )

gabrielKerekes (commented):
I just checked the public API 🙂

src/types/public.ts
jaskp: I would prefer if Ledger JS returned something at least roughly like [COSE_Sign1](https://cips.cardano.org/cip/CIP-0008#signing-and-verification)
```
COSE_Sign1 = [
    Headers,
    payload : bstr / nil,
    signature : bstr
]
```
meaning I'm in favor of returning headers. 

The payload _is_ kind of unnecessary, I guess we could consider removing that from Trezor as well.

janmazak: For ledger app... it is not well-suited to returning lots of data --- we never return more than fits in a single APDU, i.e. 255 bytes at most. It would make sense to me if it returned the address field content at least (which might be somewhat hard to come by from within hw-cli etc.). Ledgerjs can then perhaps assemble something.

Would it be possible to return complete Cose_sign1? @jaskp 

As for version, it would perhaps make sense if HW wallet returned some version for message signing, it might be useful in the future when something changes. I'll add it to ledger app.

src/interactions/signMessage.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  yield send({
+    p1: P1.STAGE_CHUNK,
+    p2: P2.UNUSED,
+    data: getChunkData(start, end),
+    expectedResponseLength: 0,
+  })
+  start = end
+
+  while (start < msgBytes.length) {
```
davidmisiak, line 79: Maybe a simple check that `hashPayload` is true could be performed here, so that we print a nice descriptive error message? I imagine it might help some dapp / sw wallet developers.

davidmisiak (commented):
Apart from the public API discussed above (mainly the `address` value), LGTM.

src/types/public.ts
jaskp: You're only really missing headers from `SignedMessageData` for it to be a subset of `COSE_Sign1`, the payload is negligible IMO. 

I think returning the version from HW wallet is a bit overkill, I only added it to Trezor Connect to keep the headers consistent. 

I agree with returning the address field from hw though.

src/types/public.ts
janmazak: I guess we should not, but perhaps it might ease someone's debugging one day.

src/types/public.ts
janmazak: I'd say full COSE_sign1 is fragile; there are optional elements we do not want to implement (though perhaps not in common use currently) etc. Ledgerjs will return only what is necessary (signature, key, address field). Returning anything else just leads to possible ambiguity (e.g. if we return algorithm, is it possible to choose a different algorithm?). 

The API and implementation are not easily extendable (e.g. to get several signatures, optional things in protected headers etc.), so it will require some rewrite anyway if we are going to be changed in the future. (This might require rewriting/updating CIP-8 as suggested by some stakeholders.)

janmazak: For Trezor, I'm not sure if it is worth it to change anything. @gabrielKerekes @refi93 @jaskp Opinions?

davidmisiak (approved):

PR #255, opened 2026-04-17
Cm/fixtures versions

carmenmmartnz:


src/parsing/transaction.ts
```diff
@@ -253,6 +189,70 @@ function parseVoterVotes(voterVotes: VoterVotes): ParsedVoterVotes {
   }
 }
 
+function parseCertificates(
```
janmazak, line 192: I do not understand what is the point of  moving this around.

PR #263, opened 2026-05-05
fix error status word mappings

janmazak:


carmenmmartnz (approved):
All changes look good to me. Everything that needed to be included is there, and the new status code mappings look consistent with the rest of the implementation.

PR #269, opened 2026-07-07
Update version for npm release

carmenmmartnz:


refi93 (approved):

refi93:
before merging I'd suggest fixing the yarn audit issues

carmenmmartnz:
Audit issues fixed

package.json
```diff
@@ -93,9 +93,13 @@
   "resolutions": {
     "brace-expansion@^5.0.2": "5.0.6",
     "chalk/ansi-regex": "<6.0.6",
+    "js-yaml": "^4.3.0",
```
refi93: Resolutions should be a last-resort solution as they are prone to cause compatibility issues down the line - depending package is sooner or later updated, resolution remains, potentially forcing some incompatible combination of packages, which can lead to hard-to-debug issues

What I think should work in this case is manually deleting entries for the vulnerable packages in `yarn.lock` and after running `yarn install` the latest compatible (and hopefully not vulnerable) versions should automatically be filled in/installed. If that fails (imho unlikely in this case), I'd try updating the depending packages in package.json, and only if that fails I'd add a resolution, but targetted to the offending top-level package (e.g. ` "@walletconnect/browser-utils@npm:1.8.0/detect-browser": "^5.3.0",`), to avoid shooting ourselves in the foot with future package updates

refi93 (changes requested):

refi93 (approved):

PR #273, opened 2026-09-21
Update deprecated dependencies

carmenmmartnz:
Develop has 6 deprecated packages and 17 audit findings. This takes both
  to zero.

  - mocha → v12, drops glob@10. Also removed the two mocha/* resolutions,
    which by now held versions back instead of lifting them.
   - eslint → v10. Leaves one stale peer warning from eslint-plugin-import, whose declared range stops at 9 despite working on 10. Holding eslint at 9 would only trade the warning for a deprecation.
  - remove unused eslint plugins and chalk.
  - shelljs forced to ^0.10 under flowgen, drops glob@7 and inflight.
  - node-hid forced to ^3.4, drops prebuild-install. Ledger still pins 2.1.2
    three years on.
  - brace-expansion resolution dropped. It pinned 5.0.6 exactly, blocking
    the patched 5.0.12 — three high-severity findings from one line.
  - six transitives refreshed with yarn up -R. Every advisory already had a
    fix inside its parent's declared range; the lockfile was just stale.
  - lodash and chalk/ansi-regex resolutions dropped too.


package.json
```diff
@@ -91,11 +91,11 @@
...
   },
   "resolutions": {
-    "brace-expansion@^5.0.2": "5.0.6",
     "chalk/ansi-regex": "<6.0.6",
-    "lodash": "4.18.0",
-    "mocha/diff": "^8.0.3",
-    "mocha/serialize-javascript": "^7.0.3"
+    "lodash": "^4.18.1",
+    "flowgen/shelljs": "^0.10.0",
+    "@ledgerhq/hw-transport-node-hid/node-hid": "^3.4.0",
```
refi93, line 92: could you clarify why these resolutions were needed?

If there are specific packages with vulnerable recursive dependencies I'd consider narrowing the resolution to them (as can be seen with `mocha/*` resolutions below) to prevent breaking packages that otherwise use a non-vulnerabale version. Alternatively, if the findings are low/moderate severity and concern only test or some other non-production functionality, I'd consider just ignoring these

carmenmmartnz: Thanks for your comment:

I removed `lodash` because both consumers already accept `4.18.1`, so the override was redundant.

I also removed `chalk/ansi-regex`: it was added for an old CVE, but `chalk` no longer depends on `ansi-regex`, so that resolution wasn't affecting anything anymore.

`flowgen/shelljs: ^0.10.0` stays because `flowgen` is still stuck on `shelljs 0.8.5`, which pulls deprecated `glob@7` and `inflight`. Since `flowgen` hasn't released since 2022 and its range can't reach `shelljs 0.9+`, the override is the only practical fix.

`@ledgerhq/hw-transport-node-hid/node-hid: ^3.4.0` (and `-noevents`) also stays. Ledger still pins `node-hid 2.1.2` exactly, so there's no parent version we can bump to avoid deprecated `prebuild-install`.

carmenmmartnz:
> ## Copilot review overview
> ### 🔵 Needs a closer look
> Major tooling upgrades and overriding Ledger’s pinned native `node-hid` dependency warrant human verification on supported platforms.
> 
> **Review effort:** Balanced **Findings:** None

Manually verified the node-hid 3.4.0 override on Ubuntu with a Ledger
Flex: `yarn run-example` and `yarn test-integration` both communicate
with the device.

